PR #59, opened 2026-07-13
feat(kata-guest): confidential workload support

CluEleSsUK:
Makes kata-guest-base production-ready for confidential agent workloads on Intel TDX. Three capabilities plus review hardening.

1. Per-workload RTMR[3] measurement (rtmr3-measurer)
An in-guest systemd daemon that measures each deployed workload's image digest into TDX RTMR[3] at runtime, dynamically, for any image, with no baked-in allowlist. This is what makes a deployment self-attesting: the register ends up reflecting exactly which image ran. It coexists with policy-monitor (allowlist enforcement) rather than replacing it; either or both may run depending on whether you want open measurement, enforcement, or both. Dedups by image digest, so container restarts, crash loops, and replicas of the same image don't double-extend the append-only register (a naive extend-per-start would make RTMR[3] depend on restart count and break verification).

2. Reproducible dm-verity root_hash
Two builds from the same inputs now produce a bit-for-bit identical root_hash. Achieved by building the rootfs offline with mkfs.ext4 -d and pinning every source of nondeterminism: FS UUID, dir-hash-seed, verity salt, and timestamps. This is what lets a verifier rebuild the image from source and recompute the launch measurement, instead of taking the build on trust. Without it the root_hash drifts build to build and independent verification is impossible.

3. Encrypted scratch disk for large images
Large workload images (for example the 2.3 GB AIO agent) unpack to a per-VM dm-crypt disk instead of the RAM tmpfs, which otherwise caps image size at guest memory. scratch-setup.service opens the disk with a random per-boot key held only in TDX-encrypted RAM, so the host never sees plaintext and never holds a key. If no scratch disk is attached the service is a no-op, so small-image deployments are unaffected. Pulls cryptsetup-bin via EXTRA_PKGS.

## Validation (live, tdx-dev-host-1)
- measurer restart-safe: a crash-looping busybox held the single-extend golden value through 3 restarts, confirming digest dedup.
- reproducible: two independent builds produced an identical root_hash, byte-identical across both partitions.
- encrypted disk: the 2.3 GB agent unpacked to the dm-crypt disk (host blkid sees only ciphertext), ran, and attested; also verified on an 8 GB config.
- every workload attestable: genuine non-debug TDX confirmed via Intel DCAP, with RTMR[3] equal to the deployed image digest. Attestation covers which image was deployed (see the scratch-disk integrity note below for the continuous-verification caveat).

## Dependencies
Requires steep https://github.com/confidential-dot-ai/steep/pull/41. The measurer reads and extends the 6.16 RTMR-extend sysfs.

## Known limitations / follow-ups (shipping with these tracked, not blockers)
- Scratch-disk integrity (tracked follow-up). The encrypted image store is dm-crypt with no integrity layer. This is not an app-swap or code-injection vector: the host holds no key so it cannot forge chosen plaintext (AES-XTS), a fresh per-boot key plus reformat kills cross-boot replay, and the image is digest-verified in-guest before unpack, which is the digest that lands in RTMR[3]. What remains: (a) the host can corrupt scratch blocks, which is a DoS, and (b) unlike the dm-verity root fs, where every block is re-verified at read against a measured root hash, the image store is verified only at unpack, not continuously at execution. So attestation covers which image was deployed, not that every byte later served off scratch still matches it. Close with dm-integrity (authenticated dm-crypt) before asserting continuous workload integrity in customer-facing claims or a security audit. Shipping now with the gap understood and bounded to DoS.
- qemu scratch wrapper (tracked improvement). scripts/kata-qemu-scratch-wrapper.sh is a shim/sidecar disk-attach. Kata has no per-sandbox scratch-disk knob, so the script wraps the qemu launch to attach the disk. It is a host-config helper and is deliberately not wired into the build. Fundamentally fine as a shim; the follow-up is a first-class attach (via the kata runtime or a CDI device) so disk lifecycle and GC are managed for us instead of by a wrapper.
- build.sh needs Docker (osbuilder) and sudo (loop devices and partitioning). CI heads-up: it will not run in a rootless or user-namespaced runner.
- Attestation trust-model items are tracked separately: client-side DCAP verification, RA-TLS binding of the app channel to the attested VM, /attest eventlog trim, and the non-public /verify endpoint.


joaosa:
recheck

AmeanAsad:

CluEleSsUK:
weirdly it's not registering the signing 🤔 I wonder if it's because I created the PR?

PR #60, opened 2026-07-13
feat(cds-attest): bind PQ sessions to mesh identity

CluEleSsUK:
## Summary

- bind every over-encryption session to the cluster's mesh identity: `report_data` is a transcript tagged with the protocol version, committing the issuing mesh CA, the exact mesh leaf, the X25519 and ML-KEM-768 session keys, and the 32-byte client nonce (most-stable fields first)
- prove possession of the committed leaf private key with a per-session ECDSA-SHA384 signature over the transcript hash; the transcript's version tag is the only domain-separation tag
- derive the channel from the verified transcript (HKDF salt), so certificate or session-key substitution derives a different channel
- read the TEE-held mesh leaf, key, and CA for every request, fail closed on missing or inconsistent credentials, and wire those files into the tls-lb chart
- drop the standalone `cds-cert.pem` discovery endpoint: the bundle embeds the exact chain committed by `report_data` (the in-cluster `/.well-known/cds-cert.pem` get-cert discovery is unchanged)
- the identity binding is the sole c8s-verify protocol (`c8s-verify/v1`): no version or binding negotiation, the attestation endpoint rejects any `binding` parameter, and the bundle carries no `binding` field; the tls-cert flow (`pq=false`) is unchanged

## Security guarantee

Previously, an attacker-operated LB with an allowed measurement could copy a victim cluster's public leaf/CA chain. The hardware evidence authenticated the attacker's session key while the independent certificate check authenticated only copied public bytes.

Acceptance now means all of the following hold:

1. hardware evidence from an allowed measurement commits the session keys, nonce, leaf hash, and CA hash;
2. the exact leaf chains to a mesh CA pinned out of band; and
3. the leaf proves possession of its private key by signing the same transcript.

Copying the public chain is therefore insufficient. Authentication is classical ECDSA; X25519 + ML-KEM-768 provides hybrid session confidentiality. This change does not claim post-quantum authentication.

Matching browser verifier: confidential-dot-ai/c8s-verify-js#10.

## Validation

- go test ./...
- Go/JavaScript cross-language transcript vector
- copied-leaf-without-key rejection
- mismatched transcript/channel rejection
- chart render coverage for the TEE-held identity mounts


joaosa:

pkg/types/verify.go
```diff
@@ -2,50 +2,61 @@ package types
...
+	Version    string          `json:"version"`      // ProtocolVersion
 	Platform   string          `json:"platform"`     // "snp" today
 	Generation string          `json:"generation"`   // AMD gen: milan|genoa|turin
 	Nonce      string          `json:"nonce"`        // echoed client nonce (b64url)
 	Evidence   json.RawMessage `json:"evidence"`     // attestation-rs SnpEvidence
-	CDSCertPEM string          `json:"cds_cert_pem"` // LB leaf + mesh CA chain (PEM)
-	// Binding names what report_data commits to. Empty is treated as
-	// BindingOverEncryption for backward compatibility with existing clients.
-	Binding string `json:"binding,omitempty"`
+	CDSCertPEM string          `json:"cds_cert_pem"` // exact mesh leaf + issuing CA committed by report_data; empty for tls-cert
```
AnomalRoil, line 54: That sounds massive xD Why not rely on an out of bound API to serve the certs that would allow us to just have the hash or identifier of the cert committed here?

AnomalRoil (approved):
lgtm

Issue #49, opened 2026-07-04
fix text logo. new website font makes text logo look too tall


gruns:
<img width="1396" height="938" alt="Image" src="https://github.com/user-attachments/assets/49979828-7f9e-47c5-b672-43b46acea70f" />

PR #63, opened 2026-07-13
feat(kms): openbao-gated secrets manager + LUKS volumes

CluEleSsUK:
- attestation-gated key brokering to an openbao instance
- support for both attested and unattested openbao


internal/helmchart/c8s/templates/secret-broker.yaml
```diff
@@ -0,0 +1,184 @@
...
+kind: ConfigMap
+metadata:
+  name: {{ include "secret-broker.fullname" . }}-policy
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "secret-broker.labels" . | nindent 4 }}
+    app.kubernetes.io/component: secret-broker
+data:
+  policy.json: |
+{{ .Values.secretBroker.releasePolicy | toJson | indent 4 }}
```
joaosa, line 18: **[critical — authorization policy is control-plane mutable]** `releasePolicy` is the root of the broker's authorization decision, but this renders it as a ConfigMap owned by the adversarial API server. An attacker can replace it with a wildcard grant and restart the Deployment; `LoadPolicy` then accepts the attacker's policy and the broker uses its OpenBao credential to release everything. `readOnly`, `immutable`, and RBAC do not help when the control plane itself is malicious. The policy must be signed/versioned, and its digest plus the broker image, arguments, and trust roots must be bound into measured HOST_DATA/initdata and enforced before workloads start. Until [c8s-old#305](https://github.com/confidential-dot-ai/c8s-old/issues/305), [#49](https://github.com/confidential-dot-ai/c8s-old/issues/49), and [#57](https://github.com/confidential-dot-ai/c8s-old/issues/57) provide that anchor, fail rendering in node-as-CVM mode.

internal/helmchart/c8s/templates/secret-broker.yaml
```diff
@@ -0,0 +1,184 @@
...
+          emptyDir:
+            medium: Memory
+            sizeLimit: 1Mi
+        - name: policy
+          configMap:
+            name: {{ include "secret-broker.fullname" . }}-policy
+        {{- if $ob.credentialSecret.name }}
+        - name: openbao-creds
+          secret:
+            secretName: {{ $ob.credentialSecret.name }}
```
joaosa: **[critical — backing-store credential is disclosed to the attacker]** A Kubernetes Secret is plaintext to the API server/etcd administrator. In this threat model the attacker can read this token or AppRole `secret_id` and call OpenBao directly, bypassing peer attestation and `releasePolicy`; mounting it instead of placing it in argv only protects process listings. Do not support `credentialSecret` in adversarial-control-plane mode. The broker needs a CVM-generated key with attested authentication, or a short-lived response-wrapped credential released only after composed evidence and scoped by a least-privilege OpenBao policy. See the [Kubernetes Secret guidance](https://kubernetes.io/docs/concepts/security/secrets-good-practices/).

internal/cmds/secretbroker/peer.go
```diff
@@ -0,0 +1,128 @@
...
+	if r.TLS == nil || len(r.TLS.PeerCertificates) == 0 {
+		return "", false
+	}
+	sum := sha256.Sum256(r.TLS.PeerCertificates[0].Raw)
+	return hex.EncodeToString(sum[:]), true
+}
+
+func workloadIDFromCert(cert *x509.Certificate) string {
+	if len(cert.DNSNames) > 0 {
+		return cert.DNSNames[0]
```
joaosa, line 147: **[critical — SAN is not an attested workload identity]** Attestation binds the certificate key to a TEE, but the control plane chooses `confidential.ai/cw` / `confidential.ai/c8s-san`. In node-as-CVM every pod shares the same node measurement, so a malicious pod can request the victim SAN, mint its own certificate, and match the victim's release rule. Certificate-fingerprint token binding does not help because the attacker legitimately mints a new certificate for the spoofed ID. Require composed evidence binding the node quote, final workload/OCI spec digest, workload ID, CSR public key, and policy version ([c8s-old#57](https://github.com/confidential-dot-ai/c8s-old/issues/57), [#49](https://github.com/confidential-dot-ai/c8s-old/issues/49)); until then reject secret release in node mode.

internal/cmds/secretbroker/cmd.go
```diff
@@ -0,0 +1,140 @@
...
+	// Broker server identity. In-cluster these are produced by the injected
+	// c8s get-cert sidecar (the shared c8s-certs tmpfs); in the demo they are a
+	// generated cert. The stock agent trusts this via its configured CA bundle.
+	flags.StringVar(&cfg.tlsCert, "tls-cert", "", "PEM server certificate the broker presents to callers (required)")
+	flags.StringVar(&cfg.tlsKey, "tls-key", "", "PEM private key for --tls-cert (required)")
+
+	// Caller (workload) verification.
+	flags.StringVar(&cfg.peerVerify, "peer-verify", peerVerifyRATLS, "how to verify caller client certs: ratls (measurement-gated, needs TEE) or ca (chain to CDS mesh CA)")
+	flags.StringVar(&cfg.clientCA, "client-ca", "", "PEM CA bundle callers' certs must chain to in --peer-verify=ca mode (the CDS mesh CA)")
+	flags.StringSliceVar(&cfg.measurements, "measurements", nil, "SHA-384 hex launch measurements accepted from callers in --peer-verify=ratls mode (empty = accept any TEE measurement, UNSAFE)")
```
joaosa, line 74: **[high — default accepts an arbitrary TEE as the victim]** RATLS mode uses `RequireAnyClientCert`; with an empty measurement set, any valid TEE can self-issue a certificate whose SAN claims the victim ID. An ID-only release rule, including the documented example, then grants the victim's paths. Because the chart defaults `measurements: []`, a warning is not fail-closed. Require a non-empty measurement or policy digest at startup, or gate an explicit dev-only `--insecure-accept-any-measurement` flag. Apply the same rule to `openbaoMeasurements`, otherwise the broker can send its backing-store credential to any attested store.

internal/helmchart/c8s/templates/validations.yaml
```diff
@@ -50,6 +50,33 @@
...
+  See docs/decisions/2026-07-09-broker-peer-verify-under-kata.md.
+*/ -}}
+{{- if and .Values.kata.enabled .Values.secretBroker.enabled (eq .Values.secretBroker.peerVerify "ratls") -}}
+{{- fail "VALIDATION_ERROR kind=broker_ratls_under_kata: secretBroker.peerVerify=ratls is inert under kata.enabled=true — the in-guest ratls-mesh terminates the broker's inbound TLS and enforces caller attestation before the broker sees the request, so the broker never gets to verify the client cert or match measurements. Pin caller measurements at the mesh (ratlsMesh.measurements) or CDS instead, and set secretBroker.peerVerify=ca (identity-only, matches the mesh-delivered peer SAN). See docs/decisions/2026-07-09-broker-peer-verify-under-kata.md." -}}
+{{- end -}}
+{{- /*
+  The LUKS-open init container the webhook injects is privileged and mounts
+  /dev host-wide (cryptsetup needs the raw block device). That is tolerable
+  only when "the host" it touches is itself inside a TEE boundary: per-pod
+  kata CVMs (kata.enabled) or node-as-CVM shapes (host attestation-api
```
joaosa, line 70: **[critical — node-as-CVM is not a per-workload security boundary]** A TEE around the whole node protects memory from the hypervisor; it does not isolate ordinary processes inside that guest from an adversarial control plane. The attacker can request these annotations, alter the final PodSpec, schedule another privileged/allowed process, and access the same node `/dev`, tmpfs, or workload memory. `attestationApi.enabled` only proves the shared node and does not make this privileged init safe. Permit this path only in a per-pod CVM, or after a measured pre-kubelet enforcer validates the full final runtime spec and cannot be disabled by Kubernetes. Otherwise fail node/base rendering and remove the corresponding node-mode break-glass guidance in `docs/pitfalls.md`.

joaosa (commented):
Security review scoped to an adversarial Kubernetes control plane and node-as-CVM. The repository threat model already treats the API server/etcd, scheduler inputs, kubelet/containerd commands, and host-provided configuration as untrusted

PR #64, opened 2026-07-13
fix(cli): pass the attestation anchor unpadded so verify works on az-…

CluEleSsUK:
…snp/az-tdx

c8s verify / c8s cds verify zero-padded the expected REPORTDATA to the 64-byte hardware field before handing it to the verifier, but the Azure vTPM platforms compare the anchor raw against the quote's extraData (48-byte SHA-384), so every az-snp target failed with "TPM nonce length mismatch: quote has 48 bytes, expected 64". Carry the anchor unpadded through all gather paths — the PROTOCOL.md contract, and what attestationclient.verifySNPEvidence already does for cluster-internal pinning — and let each platform verifier do its own shaping (bare SNP/TDX pad to the field size themselves).

Adds a regression test over real recorded az-snp evidence (vTPM nonce "challenge"). It asserts the nonce gate only until the attestation-go pin moves past v0.1.0 (see docs/GAPS.md), then tightens to full-verify


CluEleSsUK:

CluEleSsUK:
recheck

joaosa (approved):

PR #74, opened 2026-07-14
feat(cds): request-handoff client + live-cluster CA handoff probe

joaosa:
## Review Focus Areas (required)

| File / Area | Focus | Why |
|-------------|-------|-----|
| `internal/issuer/handoff_pull.go` | Critical | Client half of the attested CA transfer: signer keygen, `/attest-key` EAR, `PullHandoff`, and the retry/verdict classifier |
| `internal/issuer/handoff.go` | Critical | Typed `HandoffStatusError` + bounded read of the untrusted peer error body |
| `pkg/attestationclient/verify.go` | Critical | Fail-closed guard: a measurement pin can never be satisfied on TDX (no launch measurement surfaced), so it now rejects instead of silently skipping |
| `internal/cmds/requesthandoff/run.go` | Standard | CLI composition: RA-TLS client, JWKS provider, served-`/ca` match, exit-code mapping, sanitized output |
| `test/e2e/ca-handoff.sh` | Standard | Live-cluster probe: discovery, probe-pod spec, verdict |
| `pkg/attestclient/client.go` | Light | 8KB cap on non-2xx body reads |
| `pkg/certutil/certutil.go` | Light | `NewJSONLoggerTo` (writer-aware logger so stdout stays report-only) |

## What Changed

Adds `c8s cds request-handoff`, the client side of the existing `/handoff`
protocol, and a `make test-e2e-ca-handoff` probe that runs it in-cluster to
prove the full attested CA transfer end to end. The `/handoff` server was
already implemented and unit-tested, but nothing exercised the client path or
the transfer over a real cluster network.

Hardening that fell out of driving the client end to end:

- A measurement pin now **fails closed against a TDX peer**: the TDX verifier
  surfaces no launch measurement, so the pin was silently skipped and the
  peer's self-served JWKS became the only anchor. `VerifyEvidence` now rejects
  with `ErrMeasurementNotAllowed` until TDX measurement enforcement is wired
  through.
- Non-200s from `/ca` are typed (`HandoffStatusError`) and retried on the same
  ladder as the pull stages, so a rolling peer exits 3 (unavailable) rather
  than 2 (verification failed).
- All attestclient/attestationclient non-2xx body reads are capped at 8KB,
  matching the `/handoff` cap.
- The probe validates `--timeout`, strips control characters from error output
  (untrusted peer bodies can reach stderr), and `kget` keeps kubectl warnings
  out of parsed jsonpath output.

## Why

`/handoff` lets an attested peer pull the in-memory mesh CA, but with no client
caller it had never been driven against real TEE evidence, a real EAR from a
live CDS, the real measurement gate, or the cluster network. This ships that
client (`issuer.PullHandoff`) plus a probe that exercises it, and is the reusable
building block for CDS pull-on-startup (the follow-up PR, #75): a starting CDS
will call the same `PullHandoff` to adopt a peer's CA instead of self-generating.

The probe verifies the pulled CA matches the CA served on `GET /ca`, and its
exit codes distinguish a transient/unavailable peer (retry) from a definitive
verification failure via one `ClassifyPullError` classifier. `HandoffStatusError`
is added as a typed error so callers branch on 404-disabled vs 5xx-bootstrapping
without string-matching, and non-2xx response bodies are read through
`io.LimitReader` across the handoff and attest clients.

## Design Doc

N/A (the pull-on-startup design lands with the follow-up PR).

## Checklist

- [x] New dependencies justified (if any) — none added
- [x] Tests verify invariants, not just output format — `ClassifyPullError` table, served-CA verdict wiring (`reportFor`), typed `/ca` errors, TDX fail-closed, report-has-no-key-material
- [x] No secrets in code, logs, or error messages — the pulled CA key never reaches any output path; report struct has no key fields
- [x] For TEE code: reproducible build maintained
- [x] For crypto code: constant-time, zeroization, audited libraries only — no new crypto; reuses the existing `/handoff` envelope


internal/cmds/requesthandoff/run.go
```diff
@@ -0,0 +1,258 @@
...
+		errorf(errOut, "--peer-url must use https (RA-TLS); got scheme %q", parsed.Scheme)
+		return exitUsage
+	}
+
+	pinned, err := ratls.ParseHexMeasurementsList(cfg.measurements)
+	if err != nil {
+		errorf(errOut, "--measurements: %v", err)
+		return exitUsage
+	}
+	if len(pinned) == 0 {
```
CluEleSsUK, line 99: the implication of this is that handoff is impossible in dev systems with no measurements pinned. I think that's a fine limitation to avoid footguns

internal/cmds/requesthandoff/run.go
```diff
@@ -0,0 +1,258 @@
...
+// report is the operator-facing summary printed on stdout. It deliberately
+// has no key-typed fields.
+//
+// INVARIANT: the handed-off CA private key must never reach an output,
+// logging, or marshal path.
+type report struct {
+	CACertFingerprintSHA256 string `json:"ca_cert_fingerprint_sha256"`
+	CACertSubject           string `json:"ca_cert_subject"`
+	CACertNotAfter          string `json:"ca_cert_not_after"`
+	BundleCerts             int    `json:"bundle_certs"`
```
CluEleSsUK: odd name if it's just the count of certs

internal/issuer/handoff_pull.go
```diff
@@ -0,0 +1,193 @@
...
+	logger := cfg.Logger
+	if logger == nil {
+		logger = slog.Default()
+	}
+	interval := cfg.RetryInterval
+	if interval <= 0 {
+		interval = DefaultPullRetryInterval
+	}
+
+	signer, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
CluEleSsUK: got a little confused with all the keys but is this the CA key?
I see from https://github.com/confidential-dot-ai/c8s/blob/19d0757e055e3762198e425b9d1b432c8865ff03/internal/issuer/handoff.go#L451 that we create an ephemeral keypair for ECDH

CluEleSsUK:
a small comment on the TDX claim: the TDX verifier does surface a launch measurement, and always has. What's true is narrower: c8s never wired TDX measurement enforcement into its client, and the comment misattributes that gap to the verifier. We just can't pin RTMR specifically which makes it a bit whack. It might be worth being specific about this or documenting it somewhere and linking to it so agents don't get confused

CluEleSsUK (approved):
I just have concerns about the numbers of keys floating around and that we're not accidentally adding extras, but it may be seeing ghosts

CluEleSsUK:
also another high level comment: we removed the EAR stuff from other parts of the flow; I wonder if it's the best tool in the box for this one?

CluEleSsUK:
daddy claude doesn't agree with me: the EAR stuff means we save on a few attestations, and the keypairs do basically the same thing (and stop mitm)

joaosa:
> daddy claude doesn't agree with me: the EAR stuff means we save on a few attestations, and the keypairs do basically the same thing (and stop mitm)

yeah that was my rationale

joaosa:
@CluEleSsUK addressed feedback with [2668d3d](https://github.com/confidential-dot-ai/c8s/pull/74/commits/2668d3db8318f8904a6d52eb5843d452fa3de0cb)

PR #75, opened 2026-07-14
feat(cds): pull-on-startup CA adoption + active/standby rollout (#18)

joaosa:
> Stacked on #74 (`feat/ca-handoff-probe`). Review that first; this PR's diff
> is only the pull-on-startup + topology changes on top of it.

## Review Focus Areas (required)

| File / Area | Focus | Why |
|-------------|-------|-----|
| `internal/issuer/ca_provision.go` | Critical | The generate-vs-adopt-vs-fail-closed policy at CDS startup; a wrong branch mints or mis-trusts the mesh CA |
| `internal/cmds/cds/run.go` | Critical | `ProvisionCA` replaces the unconditional self-generate; downstream CA consumers must be unchanged |
| `internal/helmchart/c8s/templates/cds.yaml` | Standard | Strategy wiring + render guards (peerUrl needs handoff.enabled + https, forbids persistence; replicas fixed at 1) + adoption startupProbe |
| `internal/helmchart/c8s/templates/_helpers.tpl` | Standard | `self` sentinel resolves to the CDS Service URL |
| `docs/decisions/2026-07-14-cds-active-active-ear-jwks.md` | Light | Stage-3 design memo (no code); records why active/active is deferred |

## What Changed

A starting CDS can now **adopt** a surviving peer's mesh CA over attested
`/handoff` (`--handoff-peer-url`) instead of self-generating, and the chart uses
a RollingUpdate that keeps one pod serving so the replacement adopts before the
old pod retires. This preserves the mesh trust root across a CDS restart.

## Why

CDS holds its mesh CA in process memory, so every restart mints a fresh CA and
forces every workload to re-bootstrap. This closes that window:

- **Stage 1** (`issuer.ProvisionCA`): with `--handoff-peer-url` set, CDS adopts
  the peer's CA using the `PullHandoff` client from #74, and **fails closed** if
  the peer is unreachable or denies the handoff (a partition never mints a
  divergent trust root). Empty peer URL keeps the cold-start self-generate path.
- **Stage 2** (chart, active/standby): setting `cds.handoff.peerUrl` flips the
  rollout to `RollingUpdate` (`maxUnavailable: 0`/`maxSurge: 1`). The new pod
  adopts from the still-serving old pod, which is the only Ready Service endpoint
  while it starts, so no per-pod addressing is needed. Sentinel `peerUrl: self`
  resolves to the CDS Service URL. Two-phase install: `handoff.enabled=true` (+
  pinned measurements) but peerUrl empty on first install, then `helm upgrade
  --set cds.handoff.peerUrl=self` to enable adoption. Render guards keep the
  singleton invariant honest: peerUrl requires handoff.enabled (else no pod
  serves /handoff and the roll wedges) and https, forbids persistence, and
  replicas is fixed at 1 (per-pod EAR keys make a second steady-state endpoint
  unsafe until stage 3). A startupProbe covers the adoption deadline so
  liveness can't kill a mid-adoption pod.
- **Stage 3** (design doc only): active/**active** (both pods serving) is blocked
  by per-pod EAR signing keys (handoff transfers the CA only, so two pods behind
  the Service have disjoint JWKS). The memo weighs sharing the EAR key (breaks on
  rotation) vs aggregating JWKS (recommended) and defers it to #75.

Proven end to end on a node-as-CVM SNP cluster: a `rollout restart` with adoption
on produced a new pod logging `source=adopted-from-peer` with the **same** CA
fingerprint as before the restart (a plain restart changes it).

## Hardening that fell out of driving adoption end to end

- The peer-JWKS refresher is scoped to the pull; it no longer re-attests the
  Service URL every minute for the process lifetime.
- Adoption refuses a handoff carrying a parent cert or a multi-cert bundle: it
  serves a single self-signed root and must not silently drop trust material.
- The fail-closed startup error names the recovery (unset `--handoff-peer-url`
  for a deliberate re-bootstrap), and operator.md documents the involuntary-loss
  runbook: if the sole pod dies (crash/OOM/node failure) no peer survives, the
  replacement crash-loops by design, and recovery is a deliberate re-bootstrap.

## Design Doc

`docs/decisions/2026-07-14-cds-active-active-ear-jwks.md` (stage 3, in this PR).

## Checklist

- [x] New dependencies justified (if any) — none added
- [x] Tests verify invariants, not just output format — generate/adopt/fail-closed policy table + chart render guards
- [x] No secrets in code, logs, or error messages — the adopted CA key never reaches any output path
- [x] For TEE code: reproducible build maintained
- [x] For crypto code: constant-time, zeroization, audited libraries only — no new crypto; reuses the `/handoff` client from #74


CluEleSsUK:
some claude thoughts:
```
  1. Adoption preserves the CA but never renews it. The adopted cert keeps its original NotAfter, CDS's readiness gates on remaining CA lifetime, and scheduled rotation is still
  unwired (CARotator has no caller). So a long-lived adoption chain marches toward a hard expiry wall that forces the very re-bootstrap this feature avoids. Worth asking what the
  intended story is (probably "rotation lands later," but it's now load-bearing).
  2. TDX: adoption inherits #74's fail-closed pin, so peerUrl is SNP-only in practice — nothing says so in values.yaml or the operator doc.
  3. Same dangling reference problem: the memos and GAPS.md "track #75", which is the PR itself, not an issue (the repo has no issues) — fine as self-reference, but #18 in #74
  still points nowhere.
  4. Minor: the allowlist-persistence exclusion is enforced and documented, but operators who currently rely on persistence.enabled for dynamic allowlist entries lose that when
  opting into adoption — deserves a louder call-out than the one-line note.
  ```
  
I think the allowlist persistence one is actually quite a pain - allowlist becoming unavailable or resetting could tank a cluster.  The old CDS could send the new CDS the allowlist and they could both persist it; there's a race condition, but better to retry allowlist add nuke everything.

File: docs/decisions/2026-07-14-cds-active-active-ear-jwks.md [hunk 1]
```diff
@@ -0,0 +1,151 @@
...
+(`keyprovider.go:89-99`) already makes this work for external consumers **once
+every pod serves the full union**: a load-balanced re-fetch lands on some pod,
+and any pod can resolve any kid.
+
+**What's missing today:** the rotator has no API to inject a foreign public key
+(`rebuildJWKS` only serves keys it generated, `rotator.go:191-205`), and the
+inbound verifier is the rotator, which rejects sibling kids. So Option B needs:
+1. A public-key ingest path on the rotator (an "external keys" set merged into
+   `rebuildJWKS` and into `PublicKey(kid)` resolution).
+2. Cross-pod discovery: each pod fetches siblings' `/.well-known/jwks.json` and
```
CluEleSsUK, line 88: does every pod now host jwks.json?

internal/cmds/cds/run.go
```diff
@@ -57,14 +57,26 @@ func run(cfg config) error {
...
+		ExpectedIssuer:    cfg.earIssuerName,
+		Timeout:           cfg.handoffPeerTimeout,
+	}, slog.Default())
 	if err != nil {
-		return fmt.Errorf("generate mesh CA: %w", err)
+		return fmt.Errorf("provision mesh CA: %w", err)
 	}
-	slog.Info("generated in-memory mesh CA",
+	slog.Info("loaded in-memory mesh CA",
+		"source", map[bool]string{true: "adopted-from-peer", false: "self-generated"}[adopted],
```
CluEleSsUK: jesus that is some kind of cursed haha

internal/issuer/ca_provision.go
```diff
@@ -0,0 +1,141 @@
...
+	httpClient, err := ratls.NewVerifyingHTTPClient(pinned, cfg.AttestationApiURL)
+	if err != nil {
+		return nil, err
+	}
+
+	// The JWKS cache must outlive the pull deadline (a kid-miss refresh can
+	// run at its edge) but not the process: stop its refresher on return.
+	provCtx, cancelProv := context.WithCancel(ctx)
+	defer cancelProv()
+	keyProvider, err := NewJWKSKeyProvider(provCtx, cfg.PeerURL+"/.well-known/jwks.json", time.Minute, httpClient, logger)
```
CluEleSsUK, line 139: ah right, just CDS pods (makes more sense)

CluEleSsUK (approved):
all good updates, though I think that a malicious control plane could still steal the CA. 
Hooking into the allowlist operator-key seems like the best to avoid this.

CluEleSsUK:
and the operator-key ought to be added to the REPORT_DATA

internal/cmds/cds/run.go
joaosa: yeah this was pure evil bool dispatching

joaosa:
@CluEleSsUK feedback implemented

PR #81, opened 2026-07-15
feat(allowlist): verify attestation in-process, drop --attestation-ap…

CluEleSsUK:
…i-url

The allowlist CLI delegated evidence verification to an attestation-api /verify endpoint whose response is unsigned — a trust inversion for a tool that verifies the cluster from outside its TCB, and unusable on --kata installs where no host-side attestation-api exists. Verification now runs in-process via the new internal/localverify engine shared with `c8s verify` (measurement pin enforced uniformly, now including TDX MR_TD). In-cluster components keep their same-TCB delegated path.


CluEleSsUK:
merging this as it's boring

PR #85, opened 2026-07-16
feat(ratls): attest CDS config + per-pod workload digests via config claims

CluEleSsUK:
…claims

Config-claims: versioned digests carried on the RA-TLS cert and bound into its hardware evidence, so a verifier that pins them turns a config or workload swap into a fail-closed error. CDS attests its operator-key set and allowlist seed; workloads attest their admitted container images. Spec: docs/ratls-claims.md. Rationale (attest-and-pin over signing/gating): docs/decisions/2026-07-16-*.

- pkg/ratls: config-claims X.509 extension (OID …59888.1.3) folded into REPORT_DATA; claims-free certs unchanged. VerifyPolicy gains fail-closed pins.
- CDS: emits operator-key + seed digests; `c8s cds verify --operator-keys /--allowlist-seed` pins them, `c8s install` prints the expectation. /handoff (v2) requires byte-equal claims both ways.
- Workload digests (off by default): get-cert fetches its pod's images from a node-local broker (nri-image-policy socket / policy-monitor loopback, caller bound by peer creds), CDS gates issuance on them being allowlisted and embeds the claim. `c8s verify --workload-image` pins it.

Detection, not boot-time prevention: only pin-holders detect a swap. Rollout: the CDS serving-cert preimage changed — upgrade clients before CDS; handoff v1<->v2 fail closed.


pkg/workloadclaims/workloadclaims.go
```diff
@@ -0,0 +1,312 @@
...
+type connKey struct{}
+
+// Serve runs the broker on l (a unix listener in both deployment shapes) until
+// ctx is done. Errors from the resolver are returned to the caller as 500s —
+// get-cert fails closed on them.
+func Serve(ctx context.Context, l net.Listener, resolver Resolver) error {
+	mux := http.NewServeMux()
+	mux.HandleFunc("GET "+DigestsPath, func(w http.ResponseWriter, r *http.Request) {
+		conn, _ := r.Context().Value(connKey{}).(net.Conn)
+		containers, err := resolver.ContainersForPeer(peerPID(conn))
```
joaosa, line 252: **[HIGH]** `peerPID` only exists in `peercred_linux.go`, so this package — and everything importing it, including the `c8s` CLI — fails to build on macOS/Windows: `undefined: peerPID`. Linux-only CI won't catch it.

Fix: add a `//go:build !linux` stub returning 0 (already treated as "no peer credentials").

pkg/workloadclaims/workloadclaims.go
```diff
@@ -0,0 +1,312 @@
...
+// socket file first (so a broker restart does not fail with EADDRINUSE) and
+// chmods the socket to 0660. Caller binding is by kernel peer credentials, so
+// the mode gates reachability only, not authorization.
+func ListenUnix(socketPath string) (net.Listener, error) {
+	_ = os.Remove(socketPath)
+	l, err := net.Listen("unix", socketPath)
+	if err != nil {
+		return nil, fmt.Errorf("listen unix %s: %w", socketPath, err)
+	}
+	if err := os.Chmod(socketPath, 0o660); err != nil {
```
joaosa, line 99: **[HIGH]** Socket is `root:root 0660` (created by root nri-image-policy; fsGroup doesn't apply to hostPath), but get-cert runs as uid 65532 with no caps ⇒ `connect()` gets `EACCES`. get-cert fails closed and the chart enables this unconditionally on node-CVM, so every `cw` pod's cert issuance breaks.

Fix: chmod `0666` (authz is SO_PEERCRED, as the comment says) or chown to get-cert's gid; check the dir is `o+x` too.

docs/getcert-workload-binding.md
```diff
@@ -0,0 +1,398 @@
...
+vector is closed.
+
+**CDS re-validates the list regardless (defense in depth).** CDS never trusts
+the broker or get-cert. It treats the forwarded digest list as an untrusted
+proposal and independently checks (a) the list hashes to the evidence-bound
+claim, and (b) **every** digest is in the allowlist store. The allowlist — not
+the broker — is the invariant, so even a reporter that lied could not smuggle
+an unallowlisted image.
+
+So the worst a compromised reporter could do is make a pod attest a set of
```
joaosa: **[MEDIUM]** No "compromised reporter" is needed: any admitted workload can run its own attest flow (the attestation-api binds caller-chosen REPORT_DATA; CDS only checks list↔claim and allowlist membership) and claim the victim's allowlisted images, satisfying `c8s verify --workload-image <victim>`.

Fix: reword to "any admitted workload can assert any allowlisted image set; pins distinguish only honest workloads" and mirror in THREAT_MODEL. Real fix aligns with #57.

internal/cmds/cds/attest.go
```diff
@@ -90,7 +104,31 @@ func (h AttestHandler) HandleAttest(w http.ResponseWriter, r *http.Request) {
...
+			attestation.WriteError(w, http.StatusBadRequest, types.ErrorCodeInvalidRequest, "invalid workload_claims encoding")
+			return
+		}
+		// A stamped workload claim is downstream-verifiable only if the leaf
+		// carries hardware evidence binding it. get-cert embeds a nonce-free
+		// RA-TLS attestation extension into the CSR for exactly this, and
+		// SignCSR copies it onto the leaf. Reject a claim whose CSR lacks one:
+		// the leaf would carry config-claims that `c8s verify` cannot check
+		// against evidence (docs/ratls.md).
+		if !csrCarriesRATLSExtension(csr) {
```
joaosa: **[MEDIUM]** Presence-only check: any blob under the RA-TLS OID passes (the test uses `{0x30, 0x00}`), so a buggy/skewed client gets a leaf whose claims can never verify — found only at relying-party time.

Fix: require the embedded report's REPORT_DATA to byte-equal `SHA-384(csrPubKey ‖ sep ‖ claimsDER)` — a local compare, no attestation-api call.

pkg/workloadclaims/workloadclaims.go
joaosa: we need to have a clear expectation about running the cli on macos

pkg/operatorauth/keyset.go
```diff
@@ -0,0 +1,56 @@
...
+			return nil, fmt.Errorf("operator key %d is nil", i)
+		}
+		der, err := x509.MarshalPKIXPublicKey(pub)
+		if err != nil {
+			return nil, fmt.Errorf("marshal operator key %d: %w", i, err)
+		}
+		fp := sha256.Sum256(der)
+		fps = append(fps, fp[:])
+	}
+	sort.Slice(fps, func(i, j int) bool { return bytes.Compare(fps[i], fps[j]) < 0 })
```
AnomalRoil: `slices.SortFunc` is preferred nowadays, btw, pretty sure it'd be just:
```
	slices.SortFunc(fps, bytes.Compare)
```

internal/cmds/policymonitor/run.go
```diff
@@ -161,6 +161,7 @@ rather than a transient runtime problem.`,
 	fs.StringVar(&cfg.CDSMeasurements, "cds-measurements", "", "comma-separated SHA-384 hex launch digests CDS's RA-TLS serving cert must match (default $C8S_CDS_MEASUREMENTS)")
 	fs.StringVar(&cfg.AttestationServiceURL, "attestation-service-url", "", "local attestation-service URL for in-guest RA-TLS evidence (default $C8S_ATTESTATION_SERVICE_URL or http://127.0.0.1:8400)")
 	fs.DurationVar(&cfg.RefreshInterval, "allowlist-refresh-interval", defaultRefreshInterval, "interval to poll CDS for allowlist updates (only when --cds-url is set)")
+	fs.StringVar(&cfg.WorkloadClaimsSocketDir, "workload-claims-socket-dir", "", "guest directory to serve this pod's admitted container digests from, over a Unix socket for in-guest get-cert (default $C8S_WORKLOAD_CLAIMS_SOCKET_DIR)")
```
AnomalRoil, line 164: how's read write access to that guest directory controller? What's its source of truth? I'm a bit puzzled at how this works atm.

internal/cmds/nri-image-policy/broker.go
```diff
@@ -0,0 +1,103 @@
...
+// belong to the pod of the calling process" for the node-CVM workload-claims
+// flow (docs/ratls.md). It is fed from the same CreateContainer /
+// Synchronize events that drive enforcement, so what it vouches for is exactly
+// what was admitted. Caller identity comes from the kernel (SO_PEERCRED →
+// cgroup → container), never from the request.
+type workloadBroker struct {
+	mu         sync.RWMutex
+	containers map[string]ctrRec // containerID -> record
+	procRoot   string
+	logger     *slog.Logger
```
AnomalRoil: dat logger sure isn't used much atm

docs/GAPS.md
```diff
@@ -11,7 +11,7 @@ final security model.
...
... with a server-enforced 5-minute maximum validity, but carry no `aud`/cluster binding: clusters that pin the **same** operator key accept each other's captured tokens within that window, so pin distinct keys per cluster. The loaded key set and allowlist seed are committed to CDS's serving attestation (config-claims in the serving-cert evidence; `docs/ratls.md`): `c8s cds verify --operator-keys/--allowlist-seed` fails closed on a mismatch, pinning from the operator's own install inputs. The `/handoff` path additionally commits the canonical operator-key-set hash into requester and issuer REPORTDATA and requires an exact match, preventing operator-policy substitution between replicas; because the keys are public this commitment is not proof of operator private-key possession. This is detection by pin-holding verifiers, not boot-time enforcement — the remaining CDS startup flags are still un-attested, and verifiers that pin nothing see only what CDS claims (THREAT_MODEL §Addressable). Longer term: commit all startup policy to attested init data and move to a CA + short-lived operator certificates. See `docs/pitfalls.md` and `docs/decisions/2026-07-01-operator-cert-allowlist-write.md`.
```
AnomalRoil: Is this last sentences around "Longer term" plans still true? I feel we might have drifted from that

docs/getcert-workload-binding.md
```diff
@@ -0,0 +1,398 @@
...
+   the broker's containers into the pod's init set and main set (by the
+   init-container names the webhook passed), and `workloadclaims.Digest`
+   commits to both into a single 32-byte `workloadDigest`; `BuildConfigClaims`
+   puts it in a config-claims extension (operator-keys and seed fields left at
+   the unset sentinel — a workload attests only its own images). (See
+   "Corner 3".)
+
+4. **get-cert binds the claim into the CSR evidence.** The claims DER is folded
+   into the attestation `REPORT_DATA`:
+   `SHA-384(csrPubkey || "c8s/config-claims/v1\0" || claimsDER || challenge)`
```
AnomalRoil: see my comment in https://github.com/confidential-dot-ai/c8s-verify-js/pull/10#discussion_r3581506586

docs/getcert-workload-binding.md
```diff
@@ -0,0 +1,398 @@
...
+  the socket keeps get-cert on one compiled path in both shapes.)
+- **DNS-immunity.** A kernel path has no name-resolution step. Cluster DNS is
+  control-plane-configured, so a hostname endpoint would be redirectable
+  *regardless of what value is baked in* — baking the name buys nothing. A unix
+  socket sidesteps resolution entirely.
+- **Non-redirectability.** get-cert bakes the socket path as a compiled
+  constant (`workloadclaims.BrokerEndpoint`, in allowlisted/measured code), so
+  the control plane cannot change *where* get-cert looks — the platform supplies
+  only the socket mount, not the path. A network endpoint would be only as
+  fixed as the arg carrying it.
```
AnomalRoil, line 435: can't the operator mount a volume on that socket path and hijack it?

docs/getcert-workload-binding.md
```diff
@@ -0,0 +1,398 @@
...
+the operator key (`docs/ratls.md`), which refuses the pod, not by
+get-cert. The pattern: go over the network by name only when you can
+authenticate the endpoint's measurement (CDS); stay on the kernel-local socket
+when what you need is co-location, which attestation cannot prove (the broker).
+
+The residual left is neither DNS nor attestation: the socket file lives on a
+node path, so a malicious *allowlisted* pod that can `hostPath`-mount that
+directory read-write could swap the socket before get-cert connects. That is a
+PodSecurity / filesystem-permission hardening (the socket dir must be
+unwritable by untrusted pods; overlaps #311), not more crypto.
```
AnomalRoil: This sounds like an issue, no ?

AnomalRoil (commented):
I'm tapping out, but I have concerns about these changes and what they imply

pkg/workloadclaims/workloadclaims.go
CluEleSsUK: yeah this seems like an error - the CLI at least should run on mac: it piggybacks off authed crane, helm, and kubectl, so we should expect it to run on sysadmin machines

docs/getcert-workload-binding.md
CluEleSsUK: hmm right now the perms for writing it are gated by root though (which @joaosa highlighted as a liveness issue earlier)

docs/GAPS.md
```diff
@@ -9,10 +9,10 @@ final security model.
...
-- The c8s infrastructure images are not pinned into NRI policy by default (tracked at [#51](https://github.com/confidential-dot-ai/c8s/issues/51)).
+- Application-secret release is not implemented.
+- Per-workload measurement allowlists are not enforced at `/attest` (unimplemented; no tracking issue yet).
```
CluEleSsUK: probably this needs an issue

docs/GAPS.md
```diff
@@ -9,10 +9,10 @@ final security model.
...
+- The c8s infrastructure images are not pinned into NRI policy by default.
```
CluEleSsUK: we actually do lift them in, but I wonder if we could bake them in at build time

docs/getcert-workload-binding.md
CluEleSsUK: no - the operator has no host perms

internal/cmds/policymonitor/run.go
CluEleSsUK: it is root:root 660 managed by the nri-image-policy/policy-monitor (which have root).

pkg/operatorauth/keyset.go
CluEleSsUK: changed

internal/cmds/cds/attest.go
CluEleSsUK: indeed a bug - fixed

pkg/workloadclaims/workloadclaims.go
CluEleSsUK: also bug - fixed

docs/GAPS.md
CluEleSsUK: removed

internal/cmds/cds/attest_workloadclaims_test.go
```diff
@@ -0,0 +1,241 @@
...
+	sent := claimsDERFor(t, nil, []string{wlDigestA})
+	otherBinding := claimsDERFor(t, nil, []string{wlDigestB}) // CSR binds a different claim
+	w := postAttestClaimsWithCSR(t, h, issueChallenge(t, h), csrWithBoundClaims(t, otherBinding), sent, nil, []string{wlDigestA})
+	if w.Code != http.StatusBadRequest {
+		t.Fatalf("status %d (want 400), body=%s", w.Code, w.Body.String())
+	}
+}
+
+func TestAttest_WorkloadClaims_RejectsUnallowlistedImage(t *testing.T) {
+	mock := newMockAttestationApi(t, "deadbeef")
```
AmeanAsad, line 155: nice :)

internal/cmds/nri-image-policy/broker.go
```diff
@@ -0,0 +1,100 @@
...
+		containers: map[string]ctrRec{},
+		procRoot:   procRoot,
+	}
+}
+
+// record notes an admitted container. Injected containers (the get-cert
+// sidecar and its wait gate) are recorded too but excluded at query time by
+// name (workloadclaims.IsInjectedContainer) — the sidecar attests the app's
+// images, not its own.
+func (b *workloadBroker) record(containerID, sandboxID, name, digest string) {
```
AmeanAsad, line 39: what is sandbox id here?

docs/getcert-workload-binding.md
```diff
@@ -0,0 +1,560 @@
...
+
+The flow looks indirect on first read. It is indirect on purpose: every
+shortcut (let the pod report its own image, name the pod in the request, trust
+get-cert outright) is a forgery vector. The answers below are the quick
+version; each points at the Corner with the full argument.
+
+**Doesn't the pod already know what image it runs?** No — and a self-report
+wouldn't be trustworthy even if it did. A pod is a set of containers; a single
+container sees its own rootfs, not the registry *digest* it was pulled as, and
+nothing about its siblings. The workload identity is the pod's whole image set,
```
AmeanAsad, line 43: the isolation boundary between containers is very week though and they share the same kernel so couldn't a container learn about it's siblings?

docs/GAPS.md
CluEleSsUK: this isn't quite what I thought and prob not something we want to do long-term either

docs/getcert-workload-binding.md
```diff
@@ -0,0 +1,560 @@
+# get-cert and the workload-digest → RA-TLS cert binding
```
AmeanAsad, line 1: in this opening paragraph it's not clear that the broker is part of the NRI plugin. i actually did think "hrm why isn't this is part of NRI" but later it becomes more clear. i think we should state that early on

docs/getcert-workload-binding.md
CluEleSsUK: updated

internal/cmds/nri-image-policy/broker.go
CluEleSsUK: every pod creation gets a sandboxID from k8s - it's used to stash the details about the pod for namespaces, cgroup, etc

docs/getcert-workload-binding.md
CluEleSsUK: Within a pod, they share a network interface so could potentially tell that other things are running in the pod, though not their digest.
Within a node, not really - they'd have to break out of the container/pod, and then they'd need root access to get /proc etc

docs/getcert-workload-binding.md
CluEleSsUK: added

docs/getcert-workload-binding.md
AmeanAsad: im still not really convinced here, unless the pods are VM isolated, if they all share the same kernel on a node then id say container isolation is definitely questionable as a defense mechanism

CluEleSsUK: happy to change the text here if you'd like, though I don't see "a single container sees its own rootfs, not the registry *digest* it was pulled as, and nothing about its siblings" as a statement about security per se, more that 'get-cert can't know its own container digest nor the pod's main container's digest'

docs/THREAT_MODEL.md
```diff
@@ -146,7 +146,8 @@ rather than restating it.
...
...** on CDS is detected only by pin-holding verifiers, never prevented at boot. The operator-key list and allowlist seed are attested: CDS binds their canonical digests into its serving-cert evidence (config-claims, `docs/ratls.md`), verifiers pin them (`c8s cds verify --operator-keys/--allowlist-seed`, `ratls.VerifyPolicy`), and `/handoff` releases the mesh CA only on byte-equal claims. Residual: the **CDS pod arguments** stay host-supplied — dropping `--operator-keys` downgrades to the attested empty key set — and enforcers/in-cluster clients pin nothing (their config is host-supplied too), so between a swap-restart and the next pinned verify the cluster serves and enforces the attacker's allowlist. Revocation of op-keys is coarse (no CRL/OCSP). | control plane / host | Run pinned verifies continuously (CI gate on `c8s verify` exit codes), not only at bootstrap; expose public ingress only behind a passing verify. Move op-keys to a CA + short-lived operator certs (`x5c`), CA-based revocation. Detection by pinning verifiers — not boot-time prevention — is the accepted posture. | `docs/ratls.md` (Config-claims); pitfalls "Operator key-pinning"; GAPS §Trust model; decision 2026-07-01 |
```
AnomalRoil:
```suggestion
| A control-plane **config swap-restart** on CDS is detected only by pin-holding verifiers, never prevented at boot. The operator-key list and allowlist seed are attested: CDS binds their canonical digests into its serving-cert evidence (config-claims, `docs/ratls.md`), verifiers pin them (`c8s cds verify --operator-keys/--allowlist-seed`, `ratls.VerifyPolicy`), and `/handoff` releases the mesh CA only on byte-equal claims. Residual: the **CDS pod arguments** stay host-supplied — dropping `--operator-keys` downgrades to the attested empty key set — and enforcers/in-cluster clients pin nothing (their config is host-supplied too), so between a swap-restart and the next pinned verify the cluster serves and enforces the attacker's allowlist. Revocation of op-keys is coarse (no CRL/OCSP). | control plane / host | Run pinned verifies continuously (CI gate on `c8s verify` exit codes), not only at bootstrap; expose public ingress only behind a passing verify. Move op-keys to a CA + short-lived operator certs (`x5c`), CA-based revocation. Detection by pinning verifiers — not boot-time prevention — is the accepted posture. | `docs/ratls.md` (Config-claims); pitfalls "Operator key-pinning"; decision 2026-07-01 |
```

docs/THREAT_MODEL.md
```diff
@@ -146,7 +146,8 @@ rather than restating it.
...
+| The **workload-image pin** (`c8s verify --workload-image`) distinguishes honest workloads only. A CDS-issued leaf commits the pod's container-image digest (config-claims), but CDS binds the claim to nothing the pod actually runs: it checks only that the forwarded list hashes to the evidence-bound digest and that every image is allowlisted. Any admitted workload can run the attest flow itself (the attestation-api binds caller-chosen REPORTDATA) and assert **any allowlisted image set** — a victim workload's included — satisfying a pin against that victim. It can never assert a *non-allowlisted* image, and image integrity is untouched (everything running is still independently allowlisted). So the pin detects an honest workload drifting or a config swap, not a lying workload asserting another's identity. | admitted (allowlisted) workload | Per-workload measurement enforced at `/attest` (bind the claim to the pod's admitted/measured images). Interim: treat the pin as honest-workload detection, not identity. | `docs/getcert-workload-binding.md` (Corner 5/6); `docs/ratls.md` (Config-claims); GAPS §Trust model |
```
AnomalRoil:
```suggestion
| The **workload-image pin** (`c8s verify --workload-image`) distinguishes honest workloads only. A CDS-issued leaf commits the pod's container-image digest (config-claims), but CDS binds the claim to nothing the pod actually runs: it checks only that the forwarded list hashes to the evidence-bound digest and that every image is allowlisted. Any admitted workload can run the attest flow itself (the attestation-api binds caller-chosen REPORTDATA) and assert **any allowlisted image set** — a victim workload's included — satisfying a pin against that victim. It can never assert a *non-allowlisted* image, and image integrity is untouched (everything running is still independently allowlisted). So the pin detects an honest workload drifting or a config swap, not a lying workload asserting another's identity. | admitted (allowlisted) workload | Per-workload measurement enforced at `/attest` (bind the claim to the pod's admitted/measured images). Interim: treat the pin as honest-workload detection, not identity. | `docs/getcert-workload-binding.md` (Corner 5/6); `docs/ratls.md` (Config-claims) |
```

pkg/workloadclaims/workloadclaims.go
```diff
@@ -0,0 +1,333 @@
...
+type connKey struct{}
+
+// Serve runs the broker on l (a unix listener in both deployment shapes) until
+// ctx is done. Errors from the resolver are returned to the caller as 500s —
+// get-cert fails closed on them.
+func Serve(ctx context.Context, l net.Listener, resolver Resolver) error {
+	mux := http.NewServeMux()
+	mux.HandleFunc("GET "+DigestsPath, func(w http.ResponseWriter, r *http.Request) {
+		conn, _ := r.Context().Value(connKey{}).(net.Conn)
+		containers, err := resolver.ContainersForPeer(peerPID(conn))
```
AnomalRoil, line 252: If `cfg.Policy.EnforceExisting`  is false, then `plugin p.Synchronize` never runs `p.runSweep`.

Which in turn never runs `p.recordForBroker`, meaning the broker is only ever updated in `CreateContainer`, but if our `nri-image-policy` pod restarts (which I believe can happen haphazardly in K8s land) then it is going to have an empty broker list that never gets updated, no ?

Meaning this here would not resolved and get-cert would always fail closed, no?

AnomalRoil (changes requested):

internal/cmds/nri-image-policy/main.go
```diff
@@ -430,6 +439,22 @@ func startHealthServer(ctx context.Context, cfg healthServerConfig) error {
...
+func startWorkloadClaimsBroker(ctx context.Context, logger *slog.Logger, broker *workloadBroker, socketPath string) error {
+	l, err := workloadclaims.ListenUnix(socketPath, workloadclaims.BrokerSocketGID)
+	if err != nil {
+		return err
+	}
+	go func() {
+		logger.Info("starting workload-claims broker", "socket", socketPath)
+		if err := workloadclaims.Serve(ctx, l, broker); err != nil {
+			logger.Error("workload-claims broker error", "error", err)
+		}
```
AnomalRoil, line 453: I hope you got a graceful shutdown on `ctx` cancellation.

pkg/workloadclaims/workloadclaims.go
```diff
@@ -0,0 +1,333 @@
...
+	client := &http.Client{Timeout: timeout}
+	url := endpoint + DigestsPath
+	if path, ok := strings.CutPrefix(endpoint, "unix://"); ok {
+		client.Transport = &http.Transport{
+			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+				var d net.Dialer
+				return d.DialContext(ctx, "unix", path)
+			},
+		}
+		url = "http://workload-claims" + DigestsPath
```
AnomalRoil: what???

pkg/workloadclaims/workloadclaims.go
```diff
@@ -0,0 +1,333 @@
...
+	}
```
AnomalRoil: is this really doing what you want to do???

AnomalRoil (approved):

pkg/workloadclaims/workloadclaims.go
CluEleSsUK: fixed

pkg/workloadclaims/workloadclaims.go
CluEleSsUK: this is for test stuff, but will just remove it

internal/cmds/nri-image-policy/main.go
CluEleSsUK: yes, just one layer down in serve

pkg/workloadclaims/workloadclaims.go
CluEleSsUK: fixed

pkg/workloadclaims/workloadclaims.go
CluEleSsUK: yes but was for test stuff - removed

PR #91, opened 2026-07-17
feat(chart): require nri-image-policy on non-kata clusters

CluEleSsUK:


joaosa:
This will conflict with #88. there the node image bakes its own fail-closed nri-image-policy, so `c8s install --cvm-mode=node` sets `nriImagePolicy.enabled=false` without kata, which this guard rejects:

```
VALIDATION_ERROR kind=require_host_image_policy: nriImagePolicy.enabled=false requires kata.enabled=true
```

Neither PR renders that combination in tests, so whichever merges second breaks node-mode installs with CI still green.

I'd exempt the node shape (same key `c8s.attestationApiURL` uses), plus a test that it renders:

```
{{- if and (not .Values.kata.enabled) (not .Values.nriImagePolicy.enabled) (ne (.Values.attestationApi.cvmMode | default "baremetal") "node") -}}
```

Also:
- the comment says nri is "the ONLY enforcement on a non-kata cluster". this will not be true under cvmMode=node (baked plugin)
- docs/operator.md ~653 documents a no-digests render with `nriImagePolicy.enabled=false` and no kata, which now fails

internal/helmchart/c8s/templates/validations.yaml
```diff
@@ -50,6 +50,19 @@
 {{- fail (printf "VALIDATION_ERROR kind=enforce_host_components: kata.enabled=true requires %s =false. kata is enforcing — every workload runs as a kata CVM, where these host-side components are replaced by their in-guest counterparts in the kata-guest-base image (in-VM ratls routing, in-guest attestation-api on loopback, in-guest policy-monitor image admission). `c8s install --cvm-mode=pod` sets them for you." (join ", " .)) -}}
 {{- end -}}
 {{- end -}}
+{{- /*
+  The mirror of enforce_host_components. Where neither kata nor the node-baked
+  plugin runs, host nri-image-policy is the ONLY image-admission enforcement —
+  the attested allowlist gate. Disabling it there runs confidential workloads
+  with no image admission at all, a silent security downgrade. Two shapes carry
```
AnomalRoil, line 57: > a silent security downgrade

how silent is that for the users on the verifier side???

AnomalRoil (commented):
I'm a bit worried the operator is able to disable too many things from the charts that wouldn't be caught anywhere in the report data or on the verifier side... Right now, it seems to me it's totally possible to totally bypass all the NRI and image policies without it being caught upon verification of the report data which would mean the end users cannot make sure security is properly enforced by the operator.

But I thought we were baking these image and NRI checks in our images so that it would be part of the launch measurement and can't be dropped silently, unlike what this PR seems to imply.

What am I missing?

CluEleSsUK:
hmm yeah it's true - if the nri-image-policy is switched off, all bets are off. 
As I understand it though, if it runs once then the node enters a fail-closed state: you can't turn off its guarantees without getting root on the box and messing with it (which would be inside the CVM, and ought not be possible through the base image).
We had attested its existence when first verifying the CDS (though I'm not sure we are now), but that's not baked into the attestation report. 

Some ways to solve this:
1. bake the NRI image policy into the CDS
2. make the NRI image policy attestable itself

I don't love either tbh

internal/helmchart/c8s/templates/validations.yaml
CluEleSsUK: nri image policy is always silent to the verifier tbh

CluEleSsUK:
have you fixed your request @joaosa :

```

  ┌───────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐
  │                   joaosa's ask                    │                                                              Status                                                               │
  ├───────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Exempt the node shape in the guard                │ ✅ Guard now reads exactly their suggestion: and (not .Values.kata.enabled) (not .Values.nriImagePolicy.enabled) (ne              │
  │                                                   │ (.Values.attestationApi.cvmMode | default "baremetal") "node")                                                                    │
  ├───────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Fix the comment claiming nri is "the ONLY         │ ✅ Rewritten to carve out both kata and cvmMode=node as exempt shapes                                                             │
  │ enforcement on a non-kata cluster"                │                                                                                                                                   │
  ├───────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ docs/operator.md ~653 no-digests render now fails │ ✅ Updated (~656) to explain it renders only because default cvmMode=node bakes its own plugin; gke/aks must keep nri enabled +   │
  │                                                   │ digest-pinned                                                                                                                     │
  ├───────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ Add a test that the node shape renders            │ ✅ TestChartAllowsImagePolicyOffInNodeMode (node + nri-off → success) and TestChartRejectsImagePolicyOffOnNonKata (gke + nri-off  │
  │                                                   │ → require_host_image_policy failure); tls-lb helpers now pin nri digests                                                          │
  └───────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘


```

PR #92, opened 2026-07-17
feat(az-tdx): wire Azure-vTPM Intel TDX end to end (installer + online verifiers)

AnomalRoil:
## Summary

Azure TDX confidential VMs attest via the vTPM HCL report (a TD quote wrapped in the HCL report at `/dev/tpm0`), exactly like `az-snp` wraps an SNP report — there is no `/dev/tdx-guest` on an AKS/Azure CVM node. The `az-tdx` platform string existed but was rejected across the installer and the online RA-TLS verify paths, so a genuine TDX node could attest at the hardware layer while c8s refused its evidence. This closes that gap by treating `az-tdx` analogously to the already-supported `az-snp`.

## Changes

- **Installer** (`cmd/c8s/install.go`): stop refusing `--cvm-mode=aks --hardware-platform=tdx`. That combination now emits the Azure-vTPM TDX shape — vTPM device (`tpm=true`, no guest device) and `cds.ratlsPlatform=tdx` / `ratlsMesh.platform=tdx`, so CDS and the mesh expect the TDX family instead of parsing the TD quote as an SNP report. Flag help and the doc comment updated.
- **attestationclient** (`pkg/attestationclient/verify.go`): `VerifyEvidence` accepts `az-tdx`, sending the 48-byte vTPM nonce (like `az-snp`) and enforcing the MRTD as the launch measurement (like `tdx`).
- **ratls verify** (`pkg/ratls/verify.go`): accept `az-tdx` under the TDX TEE type in both the dispatch guard and the `teeType` assignment.
- **docs** (`docs/ratls.md`): list `az-tdx` among the envelope platforms the online verifier accepts (native `tdx` strips its `cc_eventlog`; `az-tdx` has none).

The serving/cert side already normalized `az-tdx -> tdx`, and `cds-attest` already passes `--platform` through, so the browser (`c8s-verify-js`) `az-tdx` path needs no server change.

## Testing

- `go build ./...`, `go test ./...` green. The former "az-tdx is rejected/unsupported" assertions are inverted to acceptance tests (48-byte nonce wire form, MRTD pinning, fail-closed), plus an `aks+tdx` installer case.
- Verified against **live Azure hardware**: a real `Standard_DC4es_v6` (Intel TDX) CVM's vTPM evidence verifies through this path (via attestation-go) with correct MRTD pinning and fails closed on a wrong measurement / report_data. A `Standard_DC4as_v5` (SEV-SNP) node was exercised the same way as a control.

## Not in scope

TDX measurement pinning stays MRTD-only (parity with native `tdx`); RTMR enforcement is a separate pre-existing gap (`docs/GAPS.md`).


AnomalRoil:

CluEleSsUK (approved):

pkg/attestationclient/verify_test.go
```diff
@@ -221,11 +223,32 @@ func TestVerifyEvidenceMeasurementPolicy(t *testing.T) {
 			t.Fatal("min_tcb must not be sent on the TDX path")
 		}
 	})
+
+	t.Run("az-tdx enforces MRTD surfaced as launch digest", func(t *testing.T) {
```
AmeanAsad, line 227: FYI on azure the MRTD and RTMR measurements are somewhat moot because we don't really have access to any of azure's firmware or hypervisor stack that boots up the VM. i don't believe azure publishes any reference values as well.

AmeanAsad (approved):
left a nit on testing MRTD and RTMR values for tdx but otherwise good.

PR #94, opened 2026-07-17
fix(helmchart): remove duplicated chart_test declarations

AnomalRoil:
## Summary

CI on `main` is red: `internal/helmchart/chart_test.go` contains a second, stale copy of three declarations, which breaks the test build with redeclaration errors (and consequently fails both the Build & Test and Lint jobs).

Duplicated symbols:
- `hasHostIPEnv`
- `TestChartNodeModeAttestationApiURLUsesHostIP`
- `TestChartNonNodeModeKeepsServiceDNS`

The stale copy still asserted against `cvmMode=baremetal`, which was renamed to `pod` in #90 — so it was both a compile error and out of date. This removes the duplicate and keeps the current copy that asserts against `pod`/`gke`/`aks`.

## Test plan

- `go build ./...` passes.
- `go test -race -count=1 -timeout=120s ./...` passes (exit 0).
- `gofmt -l` clean; `golangci-lint run ./internal/helmchart/...` reports 0 issues.
- The two affected tests pass, including the `pod`/`gke`/`aks` subtests.


AnomalRoil:
duplicate of #93

PR #98, opened 2026-07-20
chore(deps): dependency refresh + migrate jwx v2 to v3

AnomalRoil:
Dependency refresh as of 2026-07-20, plus a migration off the deprecated `lestrrat-go/jwx/v2`.

## Why v3 and not v4

`jwx/v2` is deprecated and no longer maintained — `go build` emits a deprecation notice on every invocation. Upstream points at v4, but v4 imports `encoding/json/v2`, which is still experiment-gated in Go 1.26: it needs `GOEXPERIMENT=jsonv2` in the Makefile, `ci.yml`, `kata-guest-base.yml`, `kernel-snapshot.yml`, both test Dockerfiles, and any bare `go test` run outside `make`. It also splits JWKS fetching into a second module (`jwx-go/jwkfetch`) now that `jwk.Cache` has left core.

v3 is not deprecated, needs no toolchain flag, and keeps the cache in core `jwk`. Worth revisiting when `encoding/json/v2` graduates or v3 is deprecated in turn.

## Changes

`internal/issuer/keyprovider.go` tracks the v3 cache API:

- `jwk.NewCache` takes an `httprc.Client` and returns an error; `Register` takes a `ctx`
- `WithMinRefreshInterval` → `WithMinInterval`; `WithRefreshWindow` is gone (the httprc client owns polling)
- `cache.Get` → `cache.Lookup`, `key.Raw(dst)` → `jwk.Export(key, dst)`

Two behaviours needed more than a literal translation:

- **`WithWaitReady(false)`** — v3's `Register` blocks on the first fetch by default and returns its error, which would make `NewJWKSKeyProvider` hard-fail against a JWKS endpoint that is not up yet. `TestNewJWKSKeyProviderInitialFetchFailureDoesNotError` pins the opposite contract.
- **`PublicKey` tolerates a failed `Lookup`** — v2's `Get` fetched on demand; v3's `Lookup` reports "not ready" until a fetch has succeeded. Returning that error directly would leave the provider permanently broken after a failed initial fetch, so it now falls through to the existing force-refresh recovery path.

`WithFetchWhitelist` is dropped: v3's `Register` ignores it, and it was already redundant — a `Cache` only ever contacts the URLs passed to `Register`, the same single URL the allowlist pinned.

`internal/issuer/handoff_bootstrap.go` used jwx solely to read an unverified `exp` claim, so it moves to `golang-jwt/jwt/v5` — already a direct dependency and already used by `eartoken.go` in the same package. Drops a duplicate JWT library and leaves `jwk` as the only jwx surface in the tree.

## Verification

`go build ./...`, `go test ./...`, and `golangci-lint run ./...` all pass with no toolchain flags. The one remaining lint finding (`api/v1alpha2/groupversion_info.go`, controller-runtime `scheme.Builder` deprecation) is pre-existing and untouched here.


internal/issuer/handoff_bootstrap.go
```diff
@@ -211,19 +210,18 @@ func (h *localHandoffBootstrap) attestKey(ctx context.Context, pubDER []byte) (s
 // of the locally provisioned EAR. Handoff peers re-validate signature, issuer,
 // and expiry via ValidateEARToken before trusting any material.
 func HandoffEARExpiry(token string) (time.Time, error) {
```
CluEleSsUK: honestly a bit off that we don't have a parsed JWT by this point

go.work
```diff
@@ -1,3 +1,3 @@
-go 1.26
+go 1.26.3
```
CluEleSsUK, line 1: i vaguely remember some reason for not doing this... something about forcing people to have a specific toolchain when building? I'm not sure - you're the go expert :p

CluEleSsUK (approved):

go.work
AnomalRoil: Yeah, that's `go mod tidy` biting me here

PR #99, opened 2026-07-21
ci: vendor the confidential-e2e lane in-repo to unblock the merge trigger

b0xtch:
## The bug (live on main)

`confidential-e2e.yml` calls a reusable workflow in **confidential-dot-ai/confidential-ci** (private). GitHub **forbids a public repo from calling a private repo's reusable workflow** — so since it merged, every post-`Docker` run fails to parse (`error parsing called workflow … workflow was not found`). The confidential e2e has not actually run on merge.

## The fix — make the lane self-contained (the-machine pattern)

Vendor the lane into c8s and call it **locally**:
- `cvm-e2e.yml` — the measured-CVM boot primitive (boot rke2-node-as-CVM on SNP metal, assert genuine `sev-snp-guest`, deliver payload, teardown + reap)
- `e2e-c8s.yml` — the c8s payload builder (install c8s in-guest + prove a cw workload gets its CDS-issued cert). Trimmed to `on: workflow_call` only — the wrapper owns triggering; the removed `push:`/`paths:` referenced confidential-ci dirs that don't exist here.
- `confidential-e2e.yml` → `uses: ./.github/workflows/e2e-c8s.yml`

## Behavior is unchanged

Same measured rke2-node CVM boot, same in-guest install (chart packaged from c8s source at the sha under test, component image tags resolved to the newest built ≤ the sha, launch measurement **enforced** via CDS, NRI image-policy in audit), same consumption proof (cw workload gets a CDS-issued cert). **No secrets** — the c8s pull chain is anonymous-public.

## Fork-safety unchanged

The wrapper's if-gate still requires a **same-repo** (`head_repository == github.repository`) `push`/`workflow_dispatch` `Docker` success on `main`/`v*`, and there is **no `pull_request` trigger** anywhere near the self-hosted confidential runner. A fork PR cannot reach it.

## Note on duplication

`cvm-e2e.yml` now lives here and in confidential-ci (which runs the nightly drift lane). This is deliberate and temporary per the matrix ADR — c8s's boot orchestration is c8s-specific (the lib repos move to a shared runner rail and won't use it). Proven mechanism: the same lane has run green many times from confidential-ci.

### DIFF
--- a/.github/workflows/cvm-e2e.yml
+++ b/.github/workflows/cvm-e2e.yml
@@ -1,0 +1,404 @@
+# ═══════════════════════════════════════════════════════════════════════════
+# THE PRIMITIVE — "run CI inside a CVM", reusable across the whole stack.
+# ═══════════════════════════════════════════════════════════════════════════
+#
+# This is the ONE generic thing: boot a measured CVM on a platform, attest it,
+# hand a caller-supplied PAYLOAD a way to run commands inside it, tear it down.
+# It knows NOTHING about c8s. c8s-install is just one payload; `cargo test
+# --features attest` is another. Consumers stay ~15 lines.
+#
+#   WHAT to test  -> the caller's `payload_b64`   (per-repo)
+#   WHERE to test -> `platform` + `flavor`        (this file's matrix knowledge)
+#   WHEN to test  -> the caller's trigger         (not our business)
+#
+# PAYLOAD CONTRACT (runs as root INSIDE the guest):
+#   given:  $K          in-guest kubectl (timeout-wrapped) — cluster flavors only
+#           $MEAS       the node's RUNTIME SNP launch measurement (configfs-tsm),
+#                       i.e. the exact value CDS/verifiers compare against
+#           $GHCR_USER / $GHCR_TOKEN   registry creds, if the caller passed them
+#   must:   echo "@@PAYLOAD_OK@@" on success (that is the gate)
+#   may:    echo "@@ANYTHING@@" for progress — relayed to the run log live
+#
+# Everything the payload prints is streamed to the serial console and read back
+# host-side from the guest-console-log sidecar (fast); the console is used ONLY
+# to deliver + launch, never to poll.
+name: cvm-e2e (primitive)
+on:
+  workflow_call:
+    inputs:
+      platform:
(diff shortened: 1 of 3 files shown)


b0xtch:
### What this unblocks

c8s's `confidential-e2e` merge-trigger was failing to resolve — a public repo can't `uses:` a private repo's (confidential-ci) reusable workflow. This **vendors the lane into c8s** and calls it locally (`uses: ./`). Fork-safety unchanged (no `pull_request`, same-repo if-gate); no secrets in the payload; the vendored `e2e-c8s.yml` body is byte-identical to confidential-ci's copy.

### Why it's interim-pinned (important for reviewers)

The gate currently pins a **known-good bundle** — `c8s_ref=70aea72` + `attestation-api=sha-63a97d3` — and does **not** test the merge head yet. `main`'s `--cvm-mode=node` (#88) dials a node-baked attestation-api on `$(HOST_IP):8400`, which the current staged RKE2 node image doesn't bake, so `main` reds there until a baked-AA node image (confos `c8s`-profile) is built + staged. That's expected `#88` behavior, **not a c8s bug**. Old c8s (`70aea72`, pre-#88) used the chart-AA `.svc` URL served by the deployed pod → green on the same image.

So today this proves *"the pipeline + a consistent bundle stay green,"* not *"this merge integrates."*

### Flip to real per-merge testing

One-line wrapper revert in `confidential-e2e.yml`:

```yaml
c8s_ref: ${{ inputs.c8s_ref || github.event.workflow_run.head_sha }}   # drop aa_ref
```

once the baked-AA node image is staged. The `INTERIM` comment in the wrapper spells this out inline.

.github/workflows/cvm-e2e.yml
CluEleSsUK, line 1: imo it's nicer to extract the shell script steps into their own .sh files for shellcheck, linting, syntax highlighting etc, then you can call them in the github action yamlsteps

CluEleSsUK (approved):

.github/workflows/cvm-e2e.yml
b0xtch: agreed in principle but these are vendored 1:1 from confidential-ci (private repo), so I'll extract there and re vendor as a follow up (or just add actionlint to shellcheck the `run:` blocks) rather than diverge the copy in this unblock PR

PR #100, opened 2026-07-21
feat(ratls): expose verified peer config-claims to relying parties

CluEleSsUK:
A relying party could pin a config-claim (reject on mismatch) but could not read a verified peer's claims off a live connection: VerifyResult carried no claims and the mTLS verify callback discarded its result. Reading the digest was wired only into the c8s verify CLI.

Add VerifyResult.ConfigClaims (populated by VerifyCert) and ratls.PeerConfigClaims(*tls.ConnectionState), returning a verified peer's claims off the leaf — for an HTTP server, PeerConfigClaims(r.TLS). It reads, it does not re-verify: a completed RA-TLS handshake already covered the claims (folded into REPORTDATA on the RA-TLS path, or signed by the mesh CA on the CA path), so it is safe on any accepted connection. Trust model and honest-workload ceiling documented in docs/ratls.md.

### DIFF
--- a/docs/getcert-workload-binding.md
+++ b/docs/getcert-workload-binding.md
@@ -25,8 +25,10 @@ it to a pod, and returns that pod's admitted container image digests.
 `get-cert` hashes them into one `workloadDigest`, binds it into its CSR's
 attestation evidence, and forwards the plain list to CDS. CDS re-derives the
 hash, confirms every listed image is allowlisted, and stamps the claim onto
-the issued leaf. A relying party can then pin the workload:
-`c8s verify --workload-image sha256:…`.
+the issued leaf. A relying party can then pin the workload
+(`c8s verify --workload-image sha256:…`) or read a live mesh peer's digest off
+the connection with `ratls.PeerConfigClaims` (docs/ratls.md, "Reading a peer's
+claims").
 
 ---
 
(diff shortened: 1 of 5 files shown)


docs/getcert-workload-binding.md
```diff
@@ -25,8 +25,10 @@ it to a pod, and returns that pod's admitted container image digests.
 `get-cert` hashes them into one `workloadDigest`, binds it into its CSR's
 attestation evidence, and forwards the plain list to CDS. CDS re-derives the
 hash, confirms every listed image is allowlisted, and stamps the claim onto
-the issued leaf. A relying party can then pin the workload:
-`c8s verify --workload-image sha256:…`.
+the issued leaf. A relying party can then pin the workload
+(`c8s verify --workload-image sha256:…`) or read a live mesh peer's digest off
+the connection with `ratls.PeerConfigClaims` (docs/ratls.md, "Reading a peer's
```
AnomalRoil, line 30: Wait, are pods relying parties in there? Because we aren't exposing the raTLS side of things to the outside at the moment, are we?

To me when you say "relying party", I'm hearing the end users...

CluEleSsUK:
mostly a light helper - going to merge

PR #105, opened 2026-07-22
feat(luks): operator CLI for OpenBao-keyed volumes + PVC driver

CluEleSsUK:
`c8s luks` creates, lists, shows and destroys LUKS volumes whose passphrases
live in OpenBao, so an operator can provision encrypted storage for
confidential workloads.

The PVC driver provisions raw-block claims over kubectl rather than requiring
node access, which keeps the flow usable on managed clusters where operators
have no SSH to the nodes.

Self-contained: no c8s-internal imports, and nothing else in the tree calls
it yet.

Split out of #63.


internal/cmds/luks/create.go
```diff
@@ -0,0 +1,397 @@
...
+		passphraseEntropy int
+		output            string
+		localBackingDir   string
+		namespace         string
+		storageClass      string
+	)
+	cmd := &cobra.Command{
+		Use:   "create",
+		Short: "Provision an openbao-gated LUKS volume and emit its pod annotations",
+		Long: "create generates a strong passphrase, writes it to openbao, " +
```
CluEleSsUK, line 37: in principle we could have this part to be run in-cluster so the end user never has access to the secret, but this is the easy way forward

internal/cmds/luks/create.go
```diff
@@ -0,0 +1,397 @@
...
+		if err != nil {
+			return provisioned{}, err
+		}
+		// Always format-if-empty: nothing can luksFormat an unbound claim,
+		// so first-boot formatting in the pod is the only coherent mode.
+		return provisioned{
+			claim: claim, namespace: cfg.namespace, devToken: "pvc=" + claim,
+			mode: "format-if-empty", notes: notes,
+		}, nil
+	case "csi":
```
CluEleSsUK: we can use this for the more exotic storage types

internal/cmds/luks/pvc.go
```diff
@@ -0,0 +1,131 @@
...
+	"os/exec"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+// kubectlRun invokes kubectl (the same auth path `c8s install` relies on),
+// feeding stdin when non-empty. Package-level so tests can stub the cluster.
+var kubectlRun = func(ctx context.Context, stdin string, args ...string) (string, error) {
+	cmd := exec.CommandContext(ctx, "kubectl", args...)
```
CluEleSsUK, line 19: this needs a clear error if `kubectl` isn't available

AmeanAsad:
One root cause drives most of the findings below: create validates --workload, --name and --namespace as DNS-1123, but destroy, show and list only check non-empty. The read and destroy paths therefore accept inputs that create's validation suggests were already handled.

1. KV path traversal on destroy/show/list (medium). The KV metadata path is built with path.Join (openbao.go:45-57), which cleans but does not reject dot-dot segments. I confirmed that a name like x/../../victim/db reaches another workload's KV entry, and something like ../../../sys escapes the mount entirely. A DELETE on the metadata path is unrecoverable, and it runs with the operator token. Applying create's DNS-1123 validation to every subcommand, and rejecting dot and dot-dot segments outright, closes this.

2. Backing file path traversal to a root file delete (medium). destroy.go:117-129 resolves the image path without checking it stays under --local-dir. A name like /../../../var/lib/libvirt/images/vm1 escapes the directory, the losetup -j pre-check passes on an unattached path, and the os.Remove then runs as root on any .img suffixed target, VM disks included. Same fix as above, plus a filepath.Clean and prefix check against the resolved local dir.

3. kubectl flag injection via --namespace on destroy (medium). destroy skips validation and the value flows into the kubectl argv in pvc.go:91 and pvc.go:128. A namespace starting with two dashes, for example --token=..., is parsed as a kubectl global flag. create validates this, destroy should too.

4. --fstype is never validated (medium). The value is concatenated into the executed binary name mkfs.<fstype> (create.go:383-385), so a slash in it is a relative path execution primitive as root from the working directory. The raw value also lands in the pod annotation (create.go:164), where the webhook split re-runs the same pattern. An allowlist of ext4, xfs, btrfs fixes both sides.

5. Plain HTTP --openbao-addr is silently accepted (medium). cmd.go:44-46 and openbao.go:30-36 never check the scheme, so the privileged token and every passphrase cross the wire in cleartext. I would require https unless an explicit --allow-insecure-store dev flag is passed.

6. losetup --find --show without --nooverlap (medium). create.go:392-393: two concurrent creates can end up formatting a device that is not the fresh image.

7. Unbounded reads of OpenBao responses (medium). openbao.go:164 and openbao.go:170 read and decode without a limit, and the error body is embedded verbatim in returned errors. An io.LimitReader plus a truncated, sanitized error wrap is enough.

8. Passphrase handling leaves copies behind (medium). The bytes are converted to immutable strings in util.go:17, openbao.go:66 and openbao.go:192-197 and never zeroized. Stdin delivery itself is right; keeping everything in a []byte and zeroing it after use would close this.

9. Annotation values are built with an unescaped Sprintf (medium). create.go:164-165: a comma, equals sign or hash in --mount or the fstype injects or corrupts parsed k=v fields, including a second dev=, in the grammar the webhook split parses. --mount is currently only checked for being absolute.

Smaller items: the rollback in create.go:159 discards the deleteVolume error, so a stale cas=0 entry can permanently block a retry. Create always leaves the loop attached, so destroy's only local in-use guard always trips and trains operators into always passing --force. The default http.Client follows cross-host redirects with X-Vault-Token attached (Go only strips Authorization and Cookie). The token on argv is visible in ps (the help text warns and there is a file flag, which is good). Destroy unlinks without wiping, so the crypto-shredding-only property is worth one line in the docs. The package doc in cmd.go:8 is stale.

CluEleSsUK:
1. KV path traversal (destroy/show/list) — DNS-1123 validation now precedes every kvPath/kvMetaPath/kvListPath build, so /, . and .. can't reach another workload's KV entry or
escape the mount.

2. Backing-file path traversal — localImgPath now requires an absolute, filepath.Clean-stable --local-dir (it was never validated before) and refuses any joined path that isn't
a direct child of it (filepath.Dir(p) == dir), mirroring uninstall.go's validateSweepPath. create and both destroy consumers share the one helper.

3. kubectl flag injection via --namespace — same validation plus an unconditional validateNamespace runs before any kubectlRun. New
TestRunDestroyRejectsHostileFlagsBeforeKubectl asserts a --token=…-shaped namespace produces zero kubectl calls. (This also closes the claim-name --all vector, since
name/workload can no longer start with -.)

4. --fstype execution primitive — gated on a new internal/luksfs.Allowed (ext4|xfs) inside validateCreate, the single choke point ahead of both the mkfs.<fstype> exec and the
annotation. One note: the allowlist is ext4|xfs, not ext4/xfs/btrfs — it matches the mkfs tooling actually installed in the luks-open image and is a byte-identical package
shared with #107 so the two branches merge cleanly. Happy to add btrfs once btrfs-progs is in the image.

5. Plain-HTTP --openbao-addr — baoFlags.client() (the one construction point for all four subcommands) now parses the addr and requires https; http needs an explicit
--allow-insecure-store dev flag plus a stderr warning, other schemes are refused. Modeled on allowlist/cmd.go's existing --insecure gate.

6. losetup race — --nooverlap added.

7. Unbounded OpenBao reads — error bodies via io.LimitReader at 8 KiB, success decode through a 1 MiB LimitReader, and httpError.Error() strips control chars so a hostile body
can't forge log lines.

8. Passphrase zeroization — bytesReader no longer converts to string; generatePassphrase wipes its entropy buffer; runCreate defer-wipes the passphrase on all exit paths;
putPassphrase hand-builds the JSON body and wipes it after the request. Honest caveat: best-effort — the GC can relocate heap objects and the HTTP transport buffers the body,
so we minimize and wipe the copies we own rather than guarantee zero residue.

9. Annotation injection — --mount now rejects , = # and newlines (the annotation's structural chars); with fstype allowlisted, those were the only free-text inputs to the
annotation value.

Smaller items:
- Rollback error discarded — a failed rollback delete is now errors.Joined into the returned error with a pointer to c8s luks destroy, so a stale cas=0 entry is visible rather
than silently blocking retry.
- Loop device always in-use — fixed by narrowing the guard rather than detaching on success: the emitted annotation/hostPath point at the loop node with no re-attach at
pod-open, so detaching would dangle. Destroy now checks /sys/block/<loop>/holders — only a held device (open dm-crypt mapping) refuses without --force; attached-but-idle
proceeds; a failed read fails closed.
- Cross-host redirects — CheckRedirect refuses all redirects (with an error, not ErrUseLastResponse), so X-Vault-Token is never replayed to a redirect target.
- Crypto-shred docs — destroy's long help now states the KV passphrase delete is the crypto-shred and the .img unlink is cleanup, not a wipe.
- Stale package doc — dropped the "pvc … comes later" / "Stage 5 parser" lines (pvc ships in this PR).

Left the token-on-argv item as-is per your note (the --openbao-token-file path already covers it).

PR #106, opened 2026-07-22
docs(kms): demo runbook + hardware-free broker demo script

CluEleSsUK:
KMS_DEMO.md walks the KMS stack on a cluster. scripts/secret-broker-demo.sh
runs the broker against a real dev-mode OpenBao with no confidential hardware,
using a stock curl client in place of a Vault Agent and --peer-verify=ca,
since the measurement-gated path needs SEV-SNP or TDX.

Docs and scripts only; for demo purposes and can come after the other PRs.
Split out of #63.


AmeanAsad:
Some findings: 

1. The SAN trust claim is false (important). KMS_DEMO.md:129-131 says the SAN is trustworthy because the mesh CA chain-verified the leaf and that release is attestation-rooted, gated on the CDS-vouched identity. The SAN is caller-chosen: any pod can set the c8s-san annotation or the cw id, and CDS signs any name matching <name>.<ns>.svc. Chain verification proves CDS signed the SAN, not that the caller is entitled to it. As written, any pod in the demo namespace with cw: api receives the granted SAN and passes the release policy.

2. The Part 5 intruder beat only works because the intruder is polite (important). KMS_DEMO.md:505-510 (also the claims around lines 49 and 444-448) presents the deny-by-default intruder as stopped by identity-scoped policy and a CA-verified SAN. A real intruder just sets cw: api and gets the secret. Either add the explicit caveat that this demonstrates mechanics rather than a security boundary, or drop the beat until the identity binding lands. The workloadImages section further down already does this honestly and cites the threat model; the SAN path should meet the same standard.

3. The runbook breaks against the chart split (important). kms-demo-values.yaml never sets luks.enabled (lines 107-138), c8s install --kms does not set it either, and the operator only renders --luks-open-image when it is set. The webhook then (correctly, fail-closed) rejects the Part 8 pod at admission, so Parts 7 through 10 fail as written.

4. The cvm-mode story contradicts the chart guard (important). The doc assumes --cvm-mode aks as the primary path (lines 58-62), but the chart's luks_plain_baremetal validation requires attestationApi.cvmMode == node for LUKS, so aks and gke render-fail with LUKS enabled. The stale row at line 760 still keys on the old attestationApi.enabled condition. The doc and the guard need to agree.

5. The leftover mapper claim is wrong (medium). KMS_DEMO.md:668-670 says that after pod deletion the data is unrecoverable and the leftover is only a host-resource leak. While the mapper is open, any same-named pod adopts it without a passphrase, so the leftover is a live decrypted volume, not just a leak. The teardown PR fixes the leak; the doc should not pre-excuse it.

6. The demo script can write to the wrong store (medium). scripts/secret-broker-demo.sh:37 runs bao kv put without BAO_TOKEN=root. On a clean machine it fails; with an ambient BAO_TOKEN and something else on port 8200 it writes password=s3cr3t into a foreign store. Export the token explicitly.

Small items: the examples put --openbao-token root on node shell command lines in several places (lines 67, 398, 550, 579-580, 675) even though the flag's own help recommends --openbao-token-file; worth switching the examples and adding one line that the dev store's only in-cluster protection is a CNI-dependent NetworkPolicy. The script comment says it fetches a secret as an attested workload, but nothing in the demo is attested.

PR #107, opened 2026-07-22
feat(webhook): inject OpenBao agent + LUKS volumes into opted-in pods

CluEleSsUK:
Pods opting in with confidential.ai/secrets-inject get an init container that
authenticates to the broker and templates secrets into an in-memory volume,
gating the app until they land, plus an optional renewal sidecar.
confidential.ai/luks-<name> annotations additionally inject `c8s luks-open`
to open an OpenBao-keyed LUKS volume before the app starts.

Fail-closed: a pod that asks for either and finds the operator unconfigured is
rejected rather than started without its secrets.

The c8s image moves from distroless to debian-slim because luks-open shells
out to cryptsetup, mkfs and mount.

Inert unless the operator is given --secret-agent-image / --luks-open-image.

Split out of #63.


cmd/c8s/Dockerfile
```diff
@@ -1,4 +1,38 @@
-FROM gcr.io/distroless/static:nonroot
+# The c8s image ships every subcommand the operator + injected init
```
CluEleSsUK: maybe it's worth having 2 separate dockerfiles then for hardening

cmd/c8s/Dockerfile
```diff
@@ -1,4 +1,38 @@
...
+# containers need. Historically distroless/static:nonroot, but the LUKS
+# path (`c8s luks-open`, injected by the pod-mutator webhook) shells out
+# to `cryptsetup` + `mkfs.<fs>` + `mount`, none of which distroless
+# carries. Debian-slim gives us all three plus a stable libc and a
+# nonroot user, at ~35 MB compressed vs distroless's ~2 MB — the trade
+# is deliberate: the c8s binary itself is static, but bundling cryptsetup
+# as a subprocess is the simplest correct answer for now (a purpose-built
+# luks-open image is tracked in docs/decisions/2026-07-10-openbao-luks.md
```
CluEleSsUK: this doc doesn't exist 🤔

docs/pitfalls.md
```diff
@@ -689,3 +689,50 @@ are not regular files), so the RO mount still prevents a socket-file swap
...
+but the loop lingers until the mapper is closed. Observed live on the node-CVM/TDX demo:
+after `c8s luks destroy --force`, `/dev/mapper/c8s-data` was still open and `/dev/loop0`
+still attached to the deleted file. This is **not** a confidentiality gap — the OpenBao
+passphrase and the backing file are removed, so the ciphertext is unrecoverable; the leftover
+is a dangling mapper/loop (a host-resource leak, worst case exhausts loop devices / blocks
+re-creating a volume of the same name).
+
+**Fix — two parts:**
+
+1. *Make `destroy` self-sufficient.* In `destroyLocal` (and the in-use pre-check), close the
```
CluEleSsUK: should just implement this then

internal/webhook/pod_mutator.go
```diff
@@ -164,6 +164,25 @@ type Config struct {
...
+	// disables secrets injection.
+	SecretAgentImage string
+
+	// SecretAgentCommand is the agent binary in SecretAgentImage ("bao" for
+	// OpenBao, "vault" for HashiCorp Vault). Empty defaults to "bao".
+	SecretAgentCommand string
+
+	// SecretBrokerURL is the default secret-broker base URL the injected agent
+	// dials (a pod can override with confidential.ai/secrets-broker).
+	SecretBrokerURL string
```
CluEleSsUK, line 179: a naughty control plane could overwrite the secrets broker URL - in the unattested mode this could spell disaster

internal/webhook/secrets.go
```diff
@@ -0,0 +1,292 @@
...
+	if cfg.SecretAgentCommand != "" {
+		return cfg.SecretAgentCommand
+	}
+	return "bao"
+}
+
+// agentSecurityContext is the strict context for the c8s-owned config-render
+// init container, which runs the c8s image and needs no privileged bits.
+func agentSecurityContext() *corev1.SecurityContext {
+	f, tr := false, true
```
CluEleSsUK: jesus go sucks :p

AmeanAsad:
Findings:

1. dev= is still unvalidated, and format-if-empty still means "not LUKS", not "empty" (high). internal/webhook/luks.go:132-134 and internal/cmds/luksopen/cmd.go:135-147 are unchanged from the original PR. Any workload that can create pods can name any host device, and a device with a partition table, a filesystem and data, but no LUKS header, gets luksFormat --batch-mode. That includes the node root disk. Please gate dev= at admission against an operator allowlist or a platform-provisioned device registry (the pvc= path shows the right shape), require blkid -p to come back empty with no partition table before formatting, and consider delivering the specific device as a volumeDevice instead of mounting all of host /dev into the privileged container.

2. Mapper adoption is still open (high). luksopen/cmd.go:121-122 and 149-155: mapper names are still c8s-<annotation>, kernel-global and predictable, and openOne skips luksOpen whenever the mapper already exists, without checking with cryptsetup status that it maps the requested device. A co-located pod with the same volume name mounts another tenant's decrypted volume with no passphrase, and there is now also an mkfs path on an adopted mapper (cmd.go:159-171). Mappers are still never closed on pod deletion. Per-pod unique names (the webhook has the pod UID), a device check before any adoption, and coordination with the teardown PR would close this.

3. The new injected volumes are shadowable, and the host-dev case is worse than before (high). ensureVolume keeps a pre-existing same-named volume (pod_mutator.go:1131-1138), so a pod that pre-declares c8s-secrets, c8s-secrets-config or c8s-luks-data as a hostPath puts passphrases and the broker token sink on node disk. Pre-declaring host-dev as hostPath: / mounts the host root filesystem into the privileged LUKS container. The rebase picked up main's reserved cert volume guard, but it only covers the cert volume. Please extend that same rejection to all the new reserved names.

4. Merged alone, the feature is dead (medium). The injected init container runs c8s secret-agent-config (secrets.go:149-152), which does not exist in this binary; it lives in the broker split. On its own this PR crashloops every opted-in pod in init. Please either enforce the merge order explicitly or move the subcommand, and ideally add a guard so this cannot recur.

5. The LUKS secret= knob is parsed and required but never consumed (medium). luks.go:36-37 and 142-149: the passphrase actually comes from the secret-<name> annotation, so the documented knob is a no-op. Consume it or drop it.

6. No kata confinement check for the privileged container (medium). pod_mutator.go:727-729 and 822-825: a pod-set runtimeClassName skips kata injection but keeps LUKS injection, and the luks_plain_baremetal guard only exists in the chart split.

7. The four new injected containers are missing from ReservedInjectedNames (medium). pkg/workloadclaims/workloadclaims.go:36: they get hashed into the workload digest, which churns pins on every agent image bump and breaks the documented workloadImages flow for exactly the pods this PR creates.

8. The luks-open image is unpinned (medium). cmd/luks-open/Dockerfile:8-22 uses debian:bookworm-slim by tag plus unpinned apt packages, for a privileged image whose digest feeds measurements and allowlists. Please pin the base by digest and pin package versions. Keeping the operator itself on distroless is the right call, glad that stayed.

9. fstype is unvalidated here too (medium). luksopen/cmd.go:232-252 dispatches mkfs.<fstype> and mount -t on the annotation value; the same ext4/xfs/btrfs allowlist suggested on the CLI PR applies.

Smaller items: the passphrase is copied into strings and never zeroized, and the mounts lack nosuid,nodev (luksopen/cmd.go:243-252, 277). The secrets-dir annotation is unvalidated, so a pod can shadow arbitrary paths in its own containers (secrets.go:64-66, 132). The operator does not check at startup that --secret-broker-url is set when --secret-agent-image is (operator.go:114-117). And the luksopen tests are still parse-only: the openOne device and mapper decisions are the most dangerous code in the PR and have no tests.

I would block on 1, 2 and 3, land the fstype allowlist here and in the CLI PR.

PR #108, opened 2026-07-22
feat(secretbroker): attestation-gated secret release via OpenBao (1/2)

CluEleSsUK:
The broker sits between a confidential workload and OpenBao and releases a
secret only to a caller whose attestation it has verified. Release is scoped
two ways: to named KV fields, so a policy grants one key rather than a whole
mount; and optionally to an attested workload digest — the combined role-hash
of the caller's admitted init/main image digests, read off the leaf's
config-claims and compared byte-equal against the policy's expectation.

In ratls mode the client SAN is not trusted: identity comes from the verified
config-claims, not from a name the caller chose for itself.

Split out of #63.


internal/cmds/secretbroker/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+
+	flags.StringVar(&cfg.host, "host", "0.0.0.0", "listen host")
+	flags.IntVarP(&cfg.port, "port", "p", 8443, "listen port")
+	flags.StringVar(&cfg.logLevel, "log-level", "info", "log level: debug, info, warn, error")
+
+	// Broker server identity. In-cluster these are produced by the injected
+	// c8s get-cert sidecar (the shared c8s-certs tmpfs); in the demo they are a
+	// generated cert. The stock agent trusts this via its configured CA bundle.
+	flags.StringVar(&cfg.tlsCert, "tls-cert", "", "PEM server certificate the broker presents to callers (required)")
+	flags.StringVar(&cfg.tlsKey, "tls-key", "", "PEM private key for --tls-cert (required)")
```
CluEleSsUK: generated cert in demo?? no - it's always run in-cluster, and should always use the same get-cert machinery as every other part of the system

internal/cmds/secretbroker/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+
+	// Caller (workload) verification. CDS is the trust root: callers are
+	// verified by X.509 chain to the CDS mesh CA, and identity is read from the
+	// CDS-issued leaf (SAN + config-claims).
+	flags.StringVar(&cfg.clientCA, "client-ca", "", "PEM CA bundle callers' certs must chain to (the CDS mesh CA) (required)")
```
CluEleSsUK: like the rest of the system, everything shares a CDS. If a pod who's attested to a different CDS tries to talk to us, ratls ignores it. This isn't necessary

internal/cmds/secretbroker/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+	flags.StringVar(&cfg.attestationApiURL, "attestation-api-url", "", "attestation-api URL used to verify an attested OpenBao (--openbao-attested)")
+
+	// Release policy.
+	flags.StringVar(&cfg.policyFile, "policy", "", "path to the JSON release policy (required); deny-by-default")
```
CluEleSsUK, line 63: moved out of control plane reach in a later PR

internal/cmds/secretbroker/cmd.go
```diff
@@ -0,0 +1,124 @@
...
+	flags.DurationVar(&cfg.tokenTTL, "token-ttl", time.Hour, "lifetime of broker-minted caller tokens")
+
+	// Backing OpenBao/Vault.
+	flags.StringVar(&cfg.openbaoAddr, "openbao-addr", "", "base URL of the backing OpenBao/Vault, e.g. https://openbao:8200 (required)")
+	flags.StringVar(&cfg.openbaoCA, "openbao-ca", "", "PEM CA bundle for the OpenBao TLS endpoint (empty = system roots)")
+	flags.BoolVar(&cfg.openbaoAttested, "openbao-attested", true, "require the OpenBao endpoint to present a valid TEE attestation (RA-TLS); set false for an external/managed store")
+	flags.StringSliceVar(&cfg.openbaoMeasurements, "openbao-measurements", nil, "SHA-384 hex launch measurements accepted for an attested OpenBao (empty = accept any TEE measurement, UNSAFE)")
```
CluEleSsUK, line 70: what are the implications of this being passed here? could a malicious control plane pass in `false` and break guarantees?

internal/cmds/secretbroker/openbao.go
```diff
@@ -0,0 +1,217 @@
...
+type vaultClient struct {
+	addr  string
+	httpc *http.Client
+
+	// AppRole credentials; empty when a static token is used.
+	roleID   string
+	secretID string
+
+	mu    sync.Mutex
+	token string // current store token (static or last AppRole login)
```
CluEleSsUK: indentation is odd here

CluEleSsUK: token to be replaced with attested connection between openbao and secretbroker in a later PR

internal/cmds/secretbroker/openbao.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+	return &http.Client{
+		Timeout:   15 * time.Second,
+		Transport: &http.Transport{TLSClientConfig: tlsCfg},
+	}, nil
+}
+
+// login obtains a store token via AppRole when no token is held yet. A static
+// token (set at construction) needs no login.
+func (c *vaultClient) login(ctx context.Context) error {
```
CluEleSsUK, line 93: minor, but having multiple lock/unlocks within the same function is a code smell. It's possible that the c.token set below ends up happening out of order with another routine

internal/cmds/secretbroker/README.md
```diff
@@ -0,0 +1,155 @@
...
+the broker.
+
+```
+ workload pod (CDS mesh identity)                in TCB                    store
+ ┌───────────────────────────────┐         ┌──────────────────┐     ┌──────────────┐
+ │ app ← /vault/secrets (tmpfs)   │ mTLS    │ secret-broker    │     │ OpenBao      │
+ │ Vault Agent (UNMODIFIED) ──────┼────────►│  verify peer     │     │ (attested by │
+ │   auth=cert (the c8s cert)     │ (broker │  policy check    │────►│  default, or │
+ │   address=https://broker:8443  │  ends   │  Vault translate │     │  external)   │
+ └───────────────────────────────┘  TLS)   └──────────────────┘     └──────────────┘
```
CluEleSsUK, line 19: formatting here looks off

internal/cmds/secretbroker/README.md
```diff
@@ -0,0 +1,155 @@
...
+   `(workloadId | workloadImages)` to the KV paths they may read.
+3. **Token mint**: on `cert/login` the broker mints a short-TTL token **bound to
+   the caller's client cert** (a token cannot be replayed on a different cert).
+4. **Brokered read**: the broker authenticates to OpenBao with its own identity
+   (`--openbao-token` or AppRole) and returns the KV v2 value. The store never
+   sees the workload — only the broker.
+
+`--openbao-attested` (default `true`) requires the store to present a valid TEE
+attestation (RA-TLS); set it `false` for an external/managed store, in which
+case the broker is the documented edge of the trust boundary.
```
CluEleSsUK, line 44: I wonder if this is something that should be managed in the kms installation/initial verification and handled by ratls/the CDS, rather than being handled within the secretsbroker?

internal/cmds/secretbroker/README.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+## Status
+
+Implemented and tested: the broker, the `secret-agent-config` renderer, the
+webhook agent injection, and the Helm wiring. Validated hardware-free against a
+real OpenBao (broker + unmodified `bao agent` → templated file). Pending a TEE
+cluster: live in-cluster injection ordering. Future: scoped/least-privilege
+OpenBao tokens (the broker uses one identity today), KV v1, and dynamic/transit
+engines.
```
CluEleSsUK: can remove this

internal/cmds/secretbroker/run.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	addr := fmt.Sprintf("%s:%d", cfg.host, cfg.port)
+	srv := newHTTPServer(addr, newRouter(b, cfg.maxRequestSize), cfg)
+	srv.TLSConfig = tlsCfg
+
+	go cmdsutil.ShutdownOnDone(ctx, srv, 5*time.Second)
+
+	logStartup(cfg, addr, len(policy.Rules))
+	if err := srv.ListenAndServeTLS("", ""); err != http.ErrServerClosed {
```
CluEleSsUK, line 69: using "" for both entries here looks suspicious

AmeanAsad (commented):
Findings:

1. KV path traversal with the store token (high). handleKVRead (vault.go:122-137) concatenates mount + /data/ + path from URL-decoded segments. The policy glob treats .. as a literal segment, so a grant of secret/data/api/** matches secret/data/api/../victim/db, and the broker then proxies the un-normalized path to the store. Go's http.Client follows the store's 301 path-clean redirect with X-Vault-Token intact (openbao.go:84-88, 134). I verified this end to end with a small probe: granted api/**, requested api/../victim/db, and the broker returned the victim's secret. The same path reaches /v1/sys/*. Please reject dot, dot-dot and empty segments (and validate the mount charset) before policy evaluation, set CheckRedirect to http.ErrUseLastResponse on the store client, and add a regression test; nothing in the current tests sends a .. path.

2. The store token also leaks cross-origin on any store redirect (medium). Go strips only Authorization and Cookie on redirect, so X-Vault-Token is forwarded wherever the store points. The same CheckRedirect fix covers this.

3. Consul-template injection via the secret path and field (medium). agentcfg.go:60-68 interpolates path and field raw into a double-quoted template string, and validate() only checks Name. A quote or template-brace breakout becomes a live template action; I confirmed a payload that reads the agent token sink file, and plugin or env variants are possible too. Please reject quotes, braces and backslashes in path and field, or quote them properly.

4. Token TTL, rate limiting and session bounds (medium). The default TTL is one hour (cmd.go:61-64), there is no login rate limit, and the session map grows unboundedly within the TTL (token.go:35-87). Any cert holder can exhaust memory on the component that holds all brokered secrets. A shorter TTL, a per-identity rate limit and a hard cap with eviction would do it.

5. Plain HTTP --openbao-addr is accepted (medium). run.go:92-124 never validates the scheme, and the TLS config is silently unused under http. Store credential and every brokered secret go over the wire in cleartext when attested is false. Require https unless an explicit insecure flag.

6. The store credential is a static token or AppRole pair in an etcd-readable Secret (medium). cmd.go:71-75 and run.go:75-94: unrotated, unscoped, and the AppRole role id and secret id are co-located. Per your own threat model the control plane is untrusted, so anyone who can read etcd bypasses the broker, the policy and attestation entirely. I saw the cert-auth follow-up PR; it should land with or before this one, or the interim should be documented as explicit and temporary with a scoped token.

7. The attested store check trusts an unsigned verdict (medium). openbao.go:63-69 delegates to the attestation-api /verify endpoint, which returns unsigned verdicts over a URL that accepts plain http. A forged "verified" makes the broker hand its store credential to a fake store. Signing the verdicts, or mTLS plus https-only, closes this.

8. The server cert is loaded once and never reloaded (medium). peer.go:30-46 and run.go:62-70: LoadX509KeyPair at startup, but the mesh leaf renews (24h by default), so the broker ends up serving an expired cert and every agent handshake wedges. A GetCertificate callback or a file watch fixes it.

9. Store 404 and 403 are flattened to 502 (medium). vault.go:137-142: the status code is available on statusError but unused, so a granted-but-missing secret gets indefinite agent retries instead of a definitive 404.

Smaller items: an agent config Name of .. escapes the secrets dir by one segment via path.Join (agentcfg.go:53, 88). TLS 1.3 session resumption skips re-verification, so consider SessionTicketsDisabled given the cert-bound tokens. And the README still claims a strong, per-workload identity bind; until the issuance side binds claims to something the caller cannot choose, the README should match the threat model's ceiling, which is that the broker isolates allowlisted from not, and measured from not, but not workload from workload.

Finding 1 is the one I would block merge on as it is small and fail-closed. Findings 2 through 9 are all same-window fixes.

PR #109, opened 2026-07-22
feat(kms): deploy the secret-broker + a chart-managed dev OpenBao (2/2)

CluEleSsUK:
Chart templates for the broker, plus `c8s install --kms`, which renders an
in-chart OpenBao so the KMS path can be exercised without standing up an
external store first. Explicit secretBroker.openbao values still win; --kms
only fills the gaps.

Registers the openbao image as an external c8sComponent: it is built outside
this repo on its own version stream, so --image-tag rewriting and CI digest
resolution must skip it, while it still counts for NRI allowlist derivation.

Two validations fail closed: peerVerify=ratls under kata (inert — the in-guest
mesh has already terminated the caller's TLS and enforced attestation before
the broker sees the request), and luks.enabled on a non-confidential shape
(the injected privileged init container would otherwise land on a real host).

Inert by default: secretBroker.enabled=false. Enabling it also needs the
operator flags from the webhook secret-delivery PR.

Merge after the webhook secret-delivery PR if you intend to enable it.
Split out of #63.


cmd/c8s/render_values.go
```diff
@@ -345,6 +349,7 @@ func init() {
 	renderValuesCmd.Flags().StringVar(&installCvmMode, flagCvmMode, "", "CVM deployment shape (REQUIRED; orthogonal to --hardware-platform): pod (per-pod kata CVMs; disables host-side ratls-mesh/attestation-api/nri-image-policy) or node (generalized node-as-CVM native TEE device) or gke (GKE managed CVMs) or aks (vTPM /dev/tpm0)")
 	renderValuesCmd.Flags().StringVar(&installHardwarePlatform, flagHardwarePlatform, "sev-snp", "CPU-level TEE hardware (orthogonal to --cvm-mode): sev-snp (default, /dev/sev-guest) or tdx (Intel TDX, /dev/tdx-guest). Ignored when --cvm-mode=aks")
 	renderValuesCmd.Flags().BoolVar(&installKataDebug, "debug", false, "use the kata-guest-base DEBUG image variant (requires --cvm-mode=pod)")
+	renderValuesCmd.Flags().BoolVar(&installKMS, "kms", false, "emit the dev-mode in-chart OpenBao values (kms.enabled=true, secretBroker.enabled=true); dev/demo only")
```
CluEleSsUK: why dev mode only?

docs/pitfalls.md
```diff
@@ -689,3 +689,116 @@ are not regular files), so the RO mount still prevents a socket-file swap
...
+
+Symptom on the host: `kubectl describe pod` for any kata workload shows
+repeated `FailedCreatePodSandBox … Not authorized: url
+https://ghcr.io/v2/confidential-dot-ai/<image>/manifests/<tag>` events.
+
+Break-glass on a test cluster while a rebuild is pending: set
+`kata.guestImage.registryAuth: ""` in the install values, restart the
+`c8s-kata-deploy-image-puller` DS. That strips `agent.image_registry_auth`
+from the guest cmdline, and CDH pulls anonymously — works only while the
+target packages are unauthenticated-pullable. Production must rebuild.
```
CluEleSsUK: this can be removed - PAT removed on main

docs/pitfalls.md
```diff
@@ -689,3 +689,116 @@ are not regular files), so the RO mount still prevents a socket-file swap
...
+whole node is confidential, workload pods run as ordinary processes
+without the per-pod kata VM). The mesh, get-cert, and broker peer
+verification then run on the host without the kata-guest interception
+that seems to be biting the workload path. Full kata-CVM workload
+support is still the goal; this bullet gets deleted when the workload
+get-cert path is stable end-to-end.
+
+The end-to-end software story (broker + openbao + release policy + KV
+read via mTLS) is proven separately by
+`scratchpad/broker-openbao-e2e.sh` (adapts
```
CluEleSsUK: this file doesn't exist

docs/pitfalls.md
```diff
@@ -689,3 +689,116 @@ are not regular files), so the RO mount still prevents a socket-file swap
...
+repeated `FailedCreatePodSandBox … Not authorized: url
+https://ghcr.io/v2/confidential-dot-ai/<image>/manifests/<tag>` events.
+
+Break-glass on a test cluster while a rebuild is pending: set
+`kata.guestImage.registryAuth: ""` in the install values, restart the
+`c8s-kata-deploy-image-puller` DS. That strips `agent.image_registry_auth`
+from the guest cmdline, and CDH pulls anonymously — works only while the
+target packages are unauthenticated-pullable. Production must rebuild.
+
+## kata-workload mesh path: in-guest `get-cert` is not yet functional end-to-end
```
CluEleSsUK: we've switched to ca verification for ratls so this ought to work now?

docs/pitfalls.md
```diff
@@ -689,3 +689,116 @@ are not regular files), so the RO mount still prevents a socket-file swap
...
+without the per-pod kata VM). The mesh, get-cert, and broker peer
+verification then run on the host without the kata-guest interception
+that seems to be biting the workload path. Full kata-CVM workload
+support is still the goal; this bullet gets deleted when the workload
+get-cert path is stable end-to-end.
+
+The end-to-end software story (broker + openbao + release policy + KV
+read via mTLS) is proven separately by
+`scratchpad/broker-openbao-e2e.sh` (adapts
+`scripts/secret-broker-demo.sh` to hit the deployed openbao via
```
CluEleSsUK: not in this PR

internal/helmchart/c8s/templates/attestation-api.yaml
```diff
@@ -75,6 +75,16 @@ spec:
       annotations:
         checksum/config: {{ include "c8s.attestationApiConfig" (dict "root" .) | sha256sum }}
     spec:
+      {{- if eq .Values.attestationApi.cvmMode "node" }}
```
CluEleSsUK, line 78: are you very sure? this attestation-api has been working for a long time without this change

internal/helmchart/c8s/templates/kms-openbao-helpers.tpl
```diff
@@ -0,0 +1,69 @@
...
+{{- end -}}
+{{- end }}
+
+{{- define "secret-broker.openbaoCredKey" -}}
+{{- if .Values.secretBroker.openbao.credentialSecret.name -}}
+{{- .Values.secretBroker.openbao.credentialSecret.key -}}
+{{- else if .Values.kms.enabled -}}
+token
+{{- end -}}
+{{- end }}
```
CluEleSsUK, line 73: controllable by control plane?

internal/helmchart/c8s/values.yaml
```diff
@@ -1214,6 +1223,117 @@ tlsLb:
...
+## leaves this disabled. Requires secretBroker.enabled; conflicts with an
+## explicit secretBroker.openbao.address (see validations.yaml).
+## Runs the secretAgent.image (same openbao image/digest), so the derived
+## NRI allowlist already covers it.
+kms:
+  enabled: false
+  port: 8200
+  ## Dev root token, rendered into the <release>-openbao-dev-cred Secret and
+  ## wired to the broker when secretBroker.openbao.credentialSecret is unset.
+  devRootToken: root
```
CluEleSsUK, line 1288: I wonder if we could generate this on start and communicate it between secretsbroker and KMS instead? otherwise for external KMS we will need another process to bootstrap it on launch so it doesn't go through the control plane

internal/helmchart/c8s/templates/secret-broker.yaml
```diff
@@ -0,0 +1,197 @@
...
+      labels:
+        {{- include "secret-broker.selectorLabels" . | nindent 8 }}
+        app.kubernetes.io/component: secret-broker
+      annotations:
+        # The broker reads the release policy once at startup (LoadPolicy). Hash
+        # the policy into the pod template so a `helm upgrade` that changes
+        # secretBroker.releasePolicy rolls the Deployment and the new policy
+        # actually takes effect — otherwise the ConfigMap updates but the
+        # running broker keeps its stale in-memory policy.
+        checksum/policy: {{ .Values.secretBroker.releasePolicy | toJson | sha256sum }}
```
CluEleSsUK, line 62: this will need rethought

AmeanAsad (commented):
Findings:

1. attested=true with empty measurements renders without failing (medium). values.yaml:1263-1266 and secret-broker.yaml:111-117: the default measurements list is empty and there is no validations guard, so a broker configured to verify the store accepts any genuine TEE. The runtime warning alone is easy to miss, and an adversarial host can boot its own CVM store and receive the broker's store credential. Please fail render when attested is true and measurements is empty, with an explicit allowAnyMeasurement opt-out.

2. kms.enabled renders silently under enforcing kata (medium). I verified with helm template: the dev OpenBao runs as a plain host pod (no runtimeClassName, and the release namespace is excluded from kata enforcement) while the broker is pinned inside the TEE, so secrets leave the TEE over plain HTTP to an in-memory store with the well-known token root. The current validations only guard broker presence and address conflicts. This should fail render, or be made unmistakably dev-only.

3. An explicit external address with an http scheme is accepted with no guard (medium). kms-openbao-helpers.tpl:30-40: secretBroker.openbao.address set to an http URL with attested false renders fine, putting the store credential and all brokered secrets in cleartext on the pod network. A scheme check at render, with an explicit insecure opt-in, matches how the chart treats the nri and tls-lb URLs already.

4. Node mode ships an unauthenticated attestation API on the node IP (medium). This PR adds hostNetwork to the attestation-api under cvmMode=node (attestation-api.yaml:78-88) and makes chart-deployed node mode the default (install.go:1155-1164). That exposes unauthenticated plain-HTTP /attest, which binds caller-chosen REPORTDATA, on the routable node IP with no NetworkPolicy, and the broker then gates secret release on unsigned /verify responses fetched over http://$(HOST_IP). A NetworkPolicy and internalTrafficPolicy Local (or loopback) at minimum; signed verdicts are the real fix.

5. The broker renders with no credential and then crashloops (medium). When credentialSecret is unset, secret-broker.yaml:119-124 renders no credential args and the binary hard-fails at startup. The values comment documents the unauthenticated shape as supported, so render succeeds and the pod dies. This should fail at render with a typed marker like the other guards.

Smaller items: replicas above 1 has no sessionAffinity or guard, and the in-memory token store means tokens 403 on the other pods (the default of 1 is correct). The dev root token defaults to the literal root in a plain Opaque Secret, etcd and helm-state readable, with a CNI-dependent NetworkPolicy as the only barrier; I saw the cert-auth follow-up PR and it should cover this. The broker and dev store pods use the default service account with token automount on, while cds sets automountServiceAccountToken: false. An unknown authMethod silently falls through to the token branch. The pod-level runAsUser: 0 under kata is inert because all containers override it, but it is a footgun. The TestChartRejectsLUKSWithoutTEEShape test sets a key that does not exist (nriImagePolicy.deriveComponents; the real key is under bootstrapAllowlist), so it currently asserts nothing. And --node-baked-host-services is silently ignored with a non-node cvmMode.

PR #110, opened 2026-07-22
feat(tls-lb): enable the attestation sidecar by default

joaosa:
## What

Enable the tls-lb attestation sidecar (`/.well-known/c8s/`, c8s-verify/v1) **by default**, advertise the correct TEE on both platforms, and surface how to pin the cluster's launch measurement.

## Why

The sidecar only publishes a signed quote of its *own* TEE, so exposing it does not weaken the server — trust is established on the **verifier**, which pins the expected launch measurement. Defaulting it off left `/.well-known/c8s/*` returning 404/502 out of the box even though the nginx route exists; enabling it makes browser/CLI verification (`c8s verify`, c8s-verify-js) work on a plain install.

## Changes

- **`values.yaml`**: `tlsLb.attest.enabled` defaults `true`; comment states the server-safe / client-pins posture and that `platform`/`generation` are overridden per `--hardware-platform`.
- **`install.go`**: derives `tlsLb.attest.platform=tdx` and blanks the AMD-only `generation` from `--hardware-platform`, so TDX installs advertise correctly instead of the chart's `snp`/`genoa` default. New `--attest` flag (default `true`); `--attest=false` omits the sidecar.
- **`install.go`**: post-install hint. In node mode CDS, the ratls-mesh, and the sidecar are pods in the **same measured node**, so they share one launch measurement `M`. One unpinned `c8s verify` renders `M`, which then pins all three trust boundaries — clients (`--measurements`) and the internal mesh (`cds.measurements` / `ratlsMesh.measurements`). Install cannot compute `M` itself (it lives in the booted node image, not the install context), so the hint drives the observe-then-pin flow rather than pretending to auto-pin.

## Testing

- `go build ./cmd/c8s` + `go test ./cmd/c8s/ ./internal/cmds/cdsattest/` green.
- `TestAppendCvmModeInstallArgsSetsAttestationApiValue` updated for the new tdx `--set` args.
- Verified live on a TDX node-mode cluster via an equivalent surgical patch: `/.well-known/c8s/attestation?pq=false&nonce=…` returns a full attestation bundle (`evidence`, `cds_cert_pem`, `binding`); inference on the same tls-lb unaffected.

## Notes

- Default-on is a security-posture change; it is safe server-side but relies on clients pinning measurements (an unpinned verifier accepts any TEE — called out in the values comment and the hint).
- `platform`/`generation` chart defaults stay `snp`/`genoa`; only `c8s install` overrides them, so a raw `helm install` on TDX still needs `-f`.

### DIFF
--- a/cmd/c8s/install.go
+++ b/cmd/c8s/install.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -58,6 +59,7 @@ var (
 	installWorkloadRefs []string
 
 	installResolveDigests bool
+	installAttestEnabled  bool
 )
 
 // Flag names referenced in more than one place (registration plus a Changed()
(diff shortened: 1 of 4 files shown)


AnomalRoil (approved):
lgtm

PR #121, opened 2026-07-23
fix(webhook): reject non-memory-backed reserved cert volume collisions

AnomalRoil:
## Problem

The webhook injects the mesh cert store as a **memory-backed emptyDir** (`certsVolume`) so the sidecar's private keys never touch persistent, host-visible storage — they stay inside the TEE memory boundary. But `ensureVolume` keeps a pre-declared same-named volume instead of overwriting it:

```go
func ensureVolume(pod *corev1.Pod, v corev1.Volume) {
    for _, existing := range pod.Spec.Volumes {
        if existing.Name == v.Name {
            return // keeps the author's volume
        }
    }
    ...
}
```

So a pod that declares the reserved cert volume (`c8s-certs`) as a `hostPath`, PVC, or **disk-backed** emptyDir (`medium: ""`) gets the injected sidecar mounting *its* volume — writing private keys to persistent storage the host can read. (Audit finding M-07.)

## Fix

Add `rejectReservedCertVolume`, called from `Handle` when get-cert injection applies (mirroring the existing `rejectReservedCertContainer` guard). If the pod pre-declares the reserved cert volume as anything other than a memory-backed emptyDir, admission is denied. Omitting the volume is unchanged (the webhook injects the correct one); a correctly declared `emptyDir` with `medium: Memory` is accepted.

## Test

`TestHandleRejectsReservedCertVolumeCollision` covers hostPath (denied), disk-backed emptyDir (denied), and memory emptyDir (allowed). Verified the deny cases fail without the production change. Full `internal/webhook` race tests and `golangci-lint` are green.

### DIFF
--- a/internal/webhook/pod_mutator.go
+++ b/internal/webhook/pod_mutator.go
@@ -548,6 +548,14 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
 	}
 
 	if getCertNeeded {
+		// ensureVolume keeps a pre-declared same-named volume rather than
+		// overwriting it, so a pod that declares the reserved cert volume as a
+		// hostPath/PVC/disk-backed emptyDir would have its private keys written
+		// to persistent, host-visible storage outside the TEE memory boundary.
+		// Reject anything but the expected memory-backed emptyDir.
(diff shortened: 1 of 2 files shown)


joaosa (approved):

CluEleSsUK (approved):
wow spicy!

PR #127, opened 2026-07-23
refactor(kata)!: remove the in-guest registry-auth (ghcr-auth.json) m…

CluEleSsUK:
…echanism

Guest-pull now runs anonymously — the only mode the public c8s images need. The baked ghcr-auth.json put pull creds inside the dm-verity-measured rootfs (an anti-pattern), and the READ_PRIVATE_GHCR_TOKEN baking was already dead. Default behavior (public images) is unchanged.

BREAKING CHANGE: guest rootfs bytes change, so SNP/TDX reference values must be re-predicted. kata.guestImage.registryAuth and kata.gpu.guestImage.registryAuth are removed; private in-guest workload pulls are no longer supported.

I was sure this got removed before, but I guess not!


AmeanAsad:
@CluEleSsUK seems reasonable that at some point we might support customers private workloads which would require some sort of auth'd pull?

CluEleSsUK:
yes, though that will need to be baked into the kata image or injected somehow at launch time and is separate to the changes here

PR #131, opened 2026-07-23
fix(tls-lb): allow X-C8s-Session in default CORS headers

AnomalRoil:
Browser clients (c8s-verify-js) send `X-C8s-Session` on the `/.well-known/c8s/tunnel` request. The default `tlsLb.cors.allowHeaders` was `["Authorization", "Content-Type"]`, so with CORS enabled the tunnel preflight failed and the over-encrypted channel could not be established cross-origin — attestation and handshake worked, but `session.fetch()` was blocked.

Add `X-C8s-Session` to the default `allowHeaders` in `values.yaml` and the two template fallbacks in `tls-lb-helpers.tpl`, so `cors.enabled: true` works out of the box for the browser verifier. Adds a chart test asserting the header is present when CORS is enabled.


AmeanAsad (approved):
:shipit:

joaosa (approved):

CluEleSsUK (approved):

PR #135, opened 2026-07-24
feat(allowlist): per-workload container, argv, and path policy

CluEleSsUK:
Extend the CDS-hosted allowlist from a flat digest set to two layers: a digest floor (admitted by digest alone) and named workload entries that pin an init/main container set with per-container process and path policy.

Model and the decisions behind it:

- Policy is keyed and matched by container DIGEST, never by image reference. The reference a pod presents is chosen by the untrusted host and is not bound to the bytes that run; the digest is. Entry names and image labels are informational.

- entrypoint/cmd are tagged-union policies (deny/any/exact) over the container's effective argv: entrypoint governs argv[0], cmd governs argv[1:]. The digest already pins the image's baked ENTRYPOINT/CMD, so a policy only constrains what a pod may run for those bytes. A shared digest may appear under several entries with different argv; per-container admission is the union.

- Standalone/injected components (cds, get-cert, operator, nginx, ...) are floor digests: digest-only, no argv policy. This is what keeps get-cert's per-pod arguments and cds's own flags from being denied and bricking the platform on first boot.

- paths (read/write globs) is carried, validated, and attested but not yet consumed by any enforcer. When its secret-release enforcer lands, a grant must bind to the attested workload digest.

- Combinations ("only this init+main set runs together") are gated at CDS cert issuance, where the whole claimed set is visible; NRI and policy-monitor see containers one at a time and enforce per-container digest+argv. The issuance gate excludes floor digests (get-cert included) from both sides and requires the residual to equal one entry's set.

- Workload policy can tighten, so the in-guest/host consumers keep the floor additive (never shrinks) but swap the policy overlay wholesale gated on a monotonic version epoch, so a withheld or rolled-back CDS cannot keep a laxer policy live.

The document carries a schema tag bound into the attested seed digest, so a malformed body fails loud and a verifier pins the exact format. Operator keys, the RA-TLS transport, and the ETag/version counter are unchanged.

CLI: floor commands (list/export/diff/add/remove/upload) plus a workload namespace (derive/apply/edit/delete/get), lint, and inspect-image.
See docs/allowlist-and-capabilities.md for the full design and CLI reference.

Follow-ons (not in this branch): RTMR3-attested combinations so a combination gates container start rather than only issuance; the secret-release enforcer that consumes paths; upgrading adopted-workload bootstrap from floor digests to full workload entries.


docs/allowlist-and-capabilities.md
```diff
@@ -0,0 +1,310 @@
...
+> (`docker.io/vllm/vllm-openai:v0.6.3`) is chosen by the untrusted host and is
+> not bound to the bytes that run. The image *digest* is. So every trust decision
+> in this design keys on the digest — the reference is a label for humans, never
+> a lookup key.
+
+## The model
+
+The allowlist has two layers.
+
+- **Floor** — `digests`: a `digest -> image-label` map. An image whose digest is
```
CluEleSsUK, line 21: this allows simple usage as today where you just add a digest to the allowlist and it runs

internal/cmds/allowlist/cmd.go
AnomalRoil, line 1: Your PR says:

> CLI: floor commands (list/export/diff/add/remove/upload) plus a workload namespace (derive/apply/edit/delete/get), lint, and inspect-image. `workload derive pod` builds an entry from a live pod's pulled image digests, so an operator never hand-types a digest or an argv; policies enter only via derive or via files validated on read, so a raw "" or "*" can never become a policy.

And yet I cannot find a `workload derive` command in `newWorkloadCmd`?

pkg/allowlist/policy.go
```diff
@@ -0,0 +1,107 @@
...
+	case PolicyAny:
+		return true
+	case PolicyDeny:
+		return len(segment) == 0
+	case PolicyExact:
+		return equalStrings(segment, p.Argv)
+	default:
+		return false
+	}
+}
```
AnomalRoil: with `PolicyExact` you're doing an exact `equalStrings` comparison on `segment` and `p.Argv`? That will fail when `Argv` is multiple entires because `segment` is generated with `entrypointSegment` which takes only the first value of `Argv`, so someone who writes:

```
  {
    "schema": "c8s.allowlist/v1",
    "workloads": {
      "web": {
        "containers": [
          {
            "image": "docker.io/library/nginx:1.27",
            "digest": "sha256:0000000000000000000000000000000000000000000000000000000000000001",
            "entrypoint": {
              "policy": "exact",
              "argv": ["/docker-entrypoint.sh", "nginx"]
            },
            "cmd": { "policy": "any" }
          }
        ]
      }
    }
  }
``` 

would get rejected because argv has 2 entries and yet the entrypoint is expected to be a single entry by `equalStrings` since it starts with a
``` 
if len(a) != len(b) {
		return false
	}
``` 
we should have a more obvious rejection with clear logs when entrypoint.argv is of size > 1

internal/cmds/cds/attest.go
```diff
@@ -283,7 +289,71 @@ func (h AttestHandler) verifyWorkloadClaims(claimsDER []byte, initDigests, mainD
 			return fmt.Errorf("container image %s is not allowlisted", digest)
 		}
```
AnomalRoil, line 290: are these really meant to be in `verifyWorkloadClaims` or do they belong in a higher level `verify` function?

internal/cmds/cds/attest.go
```diff
@@ -283,7 +289,71 @@ func (h AttestHandler) verifyWorkloadClaims(claimsDER []byte, initDigests, mainD
...
 	}
-	return nil
+	doc, _, err := h.AllowlistStore.LoadAll()
+	if err != nil {
+		return fmt.Errorf("load allowlist: %w", err)
+	}
+	return enforceWorkloadCombination(doc, initDigests, mainDigests)
```
AnomalRoil, line 296: If `len(claimsDER) == 0`, which I think could happen often, this method returns nil and `enforceWorkloadCombination` is never run, is this meant to be like that ?

Does `enforceWorkloadCombination` belong to a higher level verify function or really in `verifyWorkloadClaims` ?

internal/cmds/policymonitor/cds_refresh.go
```diff
@@ -77,27 +78,36 @@ func runAllowlistRefresh(ctx context.Context, logger *slog.Logger, cfg *Config,
...
 	}
 	added := a.MergePulled(pulled)
-	if added > 0 {
-		logger.Info("allowlist refreshed from CDS", "version", resp.Version, "pulled", len(pulled), "added", added, "total", a.Size())
+
+	v, verr := strconv.ParseUint(version, 10, 64)
+	if verr != nil {
+		logger.Warn("allowlist refresh: unparseable CDS version; keeping current overlay", "version", version, "error", verr)
+		return
+	}
```
AnomalRoil, line 106: this probably needs to happen before `a.MergePulled` since the allowlist would be mutated by `refreshOnce` if you do the check here

internal/cmds/nri-image-policy/plugin.go
```diff
@@ -36,12 +36,59 @@ const (
...
+	}
+	return s.snap.Load()
+}
+
+// apply installs floor ∪ pulled at version, unless version is below the applied
+// one — an epoch rollback a withheld/rolled-back CDS must not use to loosen a
+// tightened policy. Reports whether it applied. Single-writer: only the pull
+// loop calls it, so the read-compare-store needs no lock against other writers.
+func (s *policyStore) apply(pulled *allowlist.Allowlist, version uint64) bool {
+	if cur := s.snap.Load(); cur != nil && version < cur.version {
```
AnomalRoil, line 83: Does this check survive a reboot? upon restart, `version` is 0 since `policyStore` is only a in-memory struct, no ? `newPolicyStore` doesn't set `version` to any persisted state... 

I'd recommend having a file guard or something, but that's maybe tricky since it requires persistent storage? WDYT?

internal/cmds/policymonitor/monitor.go
```diff
@@ -125,14 +128,62 @@ func runMonitor(ctx context.Context, cfg *Config) error {
...
+	return o.idx
+}
+
+// apply installs al's Index when version advances past the last applied — epoch
+// anti-rollback; a lower or equal version is ignored. The first apply always
+// installs. Reports whether it applied.
+func (o *policyOverlay) apply(al *allowlistpkg.Allowlist, version uint64) bool {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+	if o.idx != nil && version <= o.version {
```
AnomalRoil, line 172: `o.version` gets reset to 0 upon restart, right?

AnomalRoil (changes requested):
I believe keeping both ENTRYPOINT and CMD as potentially multi-entry arrays like Docker does is the way, unless that's really not supported in K8s or C8s.

For instance the Docker default ENTRYPOINT is `/bin/sh -c`, with 2 entries, but does not have a default CMD command.

But it does seem the [K8s docs](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes) say that:
> The command field corresponds to ENTRYPOINT, and the args field corresponds to CMD in some container runtimes.

and that:
> If you define args, but do not define a command, the default command is used with your new arguments

But it seems to be more of K8s thing, so maybe going with "command" and "args" instead of "entrypoint" and "cmd" might be best?

Re: version checks, I'm not sure how much work we want to put into preventing downgrades, but right now maybe having a "trust first version" seen is fine, but that probably deserve a comment and a test whenever that's the case so that it's somewhat explicit.

internal/cmds/allowlist/cmd.go
CluEleSsUK: I separated it to another branch - thought I'd removed all instances here but must have missed :'(

pkg/allowlist/policy.go
CluEleSsUK: I had envisaged that argv be `"[\"/docker-entrypoint.sh\", \"nginx\"]"` in this case but maybe this makes more sense

internal/cmds/cds/attest.go
CluEleSsUK: yes this needs to be here: the CDS checks all the claims requested for the cert against the allowlist. NRI image policy stops a non-allowlist image being started, but this stops an allowlisted image from requesting claims for a non-allowlisted image

internal/cmds/cds/attest.go
CluEleSsUK: yes: this path is taken if a `get-cert` isn't asking for any claims about its digest (which is supported for some special floor workloads like ratls-mesh). The corollary of this is that such a base container can't fetch secrets etc, because there is no policy attached.

internal/cmds/nri-image-policy/plugin.go
CluEleSsUK: yes and no: this state is ephemeral, but on restart it gets [fetched from the CDS](https://github.com/confidential-dot-ai/c8s/blob/5022a2e4faaaa0da0696337254ef5ee5d2435c53/internal/cmds/nri-image-policy/main.go#L319) which can be configured to write state to disk and make it persistent

internal/cmds/policymonitor/monitor.go
CluEleSsUK: same re: CDS sync

internal/cmds/nri-image-policy/plugin.go
CluEleSsUK: well one finesse: a rolled-back CDS is trusted again. I think we're doomed to accept the first allowlist post-restart in any case

internal/cmds/policymonitor/cds_refresh.go
CluEleSsUK: the floor merge is deliberately additive and version-independent (a union only ever grows, so a rolled-back floor can't loosen a tightening beyond the already-documented grow-only property)

AnomalRoil (approved):
lgtm

PR #137, opened 2026-07-24
ci: gate PRs on coverage; bring internal/ packages to >=80%

AnomalRoil:
## What

Adds a coverage gate to CI and brings every `internal/` package up to the new floor.

### Coverage gate
- New `scripts/coverage-gate.sh`: measures per-package statement coverage from a `-coverpkg=./...` profile, so packages with **no** test files still count as 0% instead of silently passing.
- New **Coverage gate** job in `ci.yml`:
  - fails if any package under `internal/` is below **80%** statement coverage;
  - on PRs, re-measures the base branch and fails if **total** coverage decreased.

### Tests
The 16 internal packages below the floor were brought up to it — total coverage goes from **67.9% to 84.9%**:

| package | before | after |
|---|---|---|
| internal/containerd | 16.3% | 100% |
| internal/cmds/requesthandoff | 36.1% | 90.2% |
| internal/cmds/getkubeconfig | 36.5% | 91.5% |
| internal/cmds/credrelease | 47.3% | 99.4% |
| internal/cmds/ratlsmesh | 49.0% | 88.4% |
| internal/cmds/probefile | 50.0% | 96.7% |
| internal/cmds/cdsattest | 59.9% | 93.0% |
| internal/cmds/cds | 66.3% | 91.9% |
| internal/cmds/getcert | 66.5% | 93.5% |
| internal/cmds/rtmr3measurer | 67.8% | 91.3% |
| internal/controller | 67.9% | 90.1% |
| internal/cmds/nri-image-policy | 69.3% | 92.6% |
| internal/cmds/allowlist | 71.2% | 92.9% |
| internal/cmds/verify | 75.1% | 94.4% |
| internal/allowlist | 76.8% | 91.7% |
| internal/fileutil | 77.3% | 86.4% |

All new tests are hermetic: httptest servers (incl. RA-TLS with real attested certs), an in-process fake containerd gRPC server, fake `iptables`/`ipset` shell binaries, controller-runtime fake clients, and throwaway in-test keys. No network, cluster, or TEE hardware required; the whole suite is race-detector clean.

### Non-test changes
Only behavior-preserving test seams:
- a few hardcoded paths/constructors hoisted from `const`/inline calls into package `var`s (`rtmr3measurer`, `credrelease`, `getcert`, `requesthandoff`, `ratlsmesh`);
- `controller.Run` split into `managerOptions`/`setupManager` helpers (identical wiring, error wrapping, and logs).

## Verification
- `go test -race -count=1 ./...` — all green
- `./scripts/coverage-gate.sh run && check` — all internal packages >= 80.2%, total 84.9%
- `make fmt`, `golangci-lint run ./...` — 0 issues


CluEleSsUK (approved):

AmeanAsad (commented):
@AnomalRoil i see bunch of tests here are more or less checking strings and not testing actual logic or functionality. 

i'm a bit concerned about this causing bloat but also having brittle tests that slow dev if they break easily without providing much value

AnomalRoil:
@AmeanAsad you're thinking of things like in "internal/cmds/allowlist/crane_fake_test.go"?

Yeah, I could avoid mocking external deps and maybe relax the CI pass conditions to only enforce increasing the code coverage without requiring 80% on internal packages.

AmeanAsad:
>@AmeanAsad you're thinking of things like in "internal/cmds/allowlist/crane_fake_test.go"?

Yep exactly 

>Yeah, I could avoid mocking external deps and maybe relax the CI pass conditions to only enforce increasing the code coverage without requiring 80% on internal packages.

@AnomalRoil  I think every test should earn it's place and should test some real logic. If we're adding brittle string checking tests to satisfy a coverage percentage, that's an anti-pattern imo and we should make the coverage percentage flexible there.

PR #143, opened 2026-07-27
fix(ci): publish on every push to main so the post-merge chain always fires

b0xtch:
`docker.yml` is the head of the whole post-merge chain: both **Kata guest base** and **confidential-e2e** fire off its completion via `workflow_run`. Its `push` trigger is path-gated, so a merge touching none of those paths does not merely skip an image build. It deletes that merge's measured guest image **and** its SEV-SNP integration run.

Two downstream files assert the opposite invariant in their own comments, and both have been silently false:

| file | its own comment |
|---|---|
| `kata-guest-base.yml` | "every merge to main builds to completion" |
| `confidential-e2e.yml` | "run the c8s e2e ... for every main merge" |

## Scale

Last 100 first-parent commits on main (2026-07-01 to 07-25): **25 non-source, 40 mixed, 35 source**. So **25% of merges get zero TEE coverage.**

Confirmed live, not inferred:

- `0e3a876` and `a7eb195`: **zero** Docker runs, `confidential-e2e` `workflow_run` instances **skipped**, and the only green runs on those shas came from a human pressing `workflow_dispatch`.
- GHCR agrees: `cds:0e3a876` is **404**, while `cds:3ace878`, `cds:64b6d7a` and `cds:main` are all 200.

## This is not lighting up an untested path

`retag-unchanged` has run **16 times on main, 68 legs, zero failures**. Five of those runs are the *exact shape* this change creates for a non-source merge (`has_images=false`, all five components retagged). Run `30024702633` took **18 seconds** end to end.

Better: that shape has already driven the full cascade green on the real SNP host. `45cd7a0` produced kata build `30024734222` and confidential-e2e `30024734575`, both success.

## Also fixes an unreported gap

`.github/scripts/docker-build-matrix.sh` was **not** in the path list, so a change to the script that decides the entire build matrix produced no Docker run and was never validated on main.

## Why the burst guard is required, not optional

`oras tag IMAGE:main <short-sha>` copies whatever `:main` points at *right now*. Retagging takes 20-30s; a source build takes ~3m45s. A non-source merge landing inside a source merge's build window would stamp the **pre-merge** manifest onto its own sha, and its e2e would boot a CVM on binaries predating its own tree and report green. Three such pairs exist in the last 100 merges.

The reason this must ship together with the trigger change: **the tell disappears.** Today staleness is inferable, because `resolve_tag` falls back to an ancestor sha different from HEAD. Once `:<short-sha>` always exists, a stale run looks *identical* to a correct one. Shipping the trigger change alone would trade a loud 25% gap for a silent 12% lie.

## What this PR does and does not prove

It touches `.github/workflows/docker.yml`, which is in the `pull_request` paths list, so it gets a Docker PR run. That proves the file parses and the job graph resolves. **It proves nothing about the push trigger, about `retag-unchanged` (gated on `github.event_name == 'push'`), or about the workflow_run cascade**, none of which a PR can exercise. The evidence above substitutes for that; a green check here should not be read as coverage.

`pull_request` keeps its path filter unchanged (all 10 entries), so PR cost is unaffected.

## Known follow-up, filed separately

This change puts ~25% more traffic into a hardware queue that **drops runs**. `concurrency: e2e-c8s-${{ github.ref }}` evaluates to `refs/heads/main` for every run, making it a depth-1 pending queue with eviction rather than a FIFO. That is a pre-existing bug that already silently drops source merges, and it is not mine to fix in this PR.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,14 +1,43 @@
 name: Docker
 
 on:
-  # Publish component images on pushes to main and on v* release tags. Feature
-  # branches don't publish: the kata-guest-base pipeline only builds for main /
-  # release tags (its job if: gates on head_branch == main || v*), so per-branch
-  # images had no consumer.
+  # Publish component images on EVERY push to main, and on v* release tags.
+  # Feature branches don't publish: the kata-guest-base pipeline only builds for
+  # main / release tags (its job if: gates on head_branch == main || v*), so
+  # per-branch images had no consumer.
+  #
+  # NO paths filter on the push trigger, deliberately. This workflow is the head
+  # of the whole post-merge chain: "Kata guest base" and confidential-e2e both
+  # fire off its completion via workflow_run, so a path gate here does not just
+  # skip an image build, it deletes that merge's measured guest image AND its
+  # SEV-SNP integration run. Both downstream files assert the opposite invariant
+  # in their own comments ("every merge to main builds to completion",
+  # kata-guest-base.yml; "run the c8s e2e ... for every main merge",
+  # confidential-e2e.yml) and both were silently false for 25 of the last 100
+  # merges. Verified: 0e3a876 and a7eb195 have ZERO Docker runs, their
+  # confidential-e2e workflow_run instances were skipped, and the only green
+  # runs on those shas came from a human pressing workflow_dispatch.
+  #
+  # A push that changes no buildable source is cheap AND is the correct thing to
+  # publish: paths-filter reports every component unchanged, docker-build-matrix.sh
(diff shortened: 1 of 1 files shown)


AmeanAsad:
@CluEleSsUK sanity checking here, do we want to autopublish every main merge or on releases only?

CluEleSsUK:
we do want to publish on every push to main, though it looks like the changes here actually push on every push to PRs too

CluEleSsUK:
ah I see - the finesse is that we only run docker build on main if certain relevant paths have changed. I'm fine with pushing every time

CluEleSsUK (approved):

joaosa (approved):

PR #145, opened 2026-07-27
feat: getcert-fast-rebind

CluEleSsUK:
the first cert requested by get-cert has no claims, because main containers haven't started yet. This PR re-requests a cert after the main containers have started in order to fetch the correct claims


CluEleSsUK:
gonna actually suggest we drop this and use the sandboxId instead of these claims

PR #147, opened 2026-07-27
feat(allowlist): argv-matched path grants and grant validation

CluEleSsUK:


AmeanAsad:
@CluEleSsUK plz add PR desc :)

AmeanAsad:
@CluEleSsUK we seem to introduce this concept of "grants" and wanted to make sure that this is the term we want to use going forward.

PR #151, opened 2026-07-27
feat(sandbox): broker sandbox-identity API + CDS-verified sandbox tokens on RA-TLS leaves

CluEleSsUK:
## The Plan

Create some new functionality for the NRI image plugin. It is the arbiter of everything related to containers and digests:
- the NRI image plugin hosts an API  `/sandbox` that allows a container to tell which sandboxId it's running in
- the NRI image plugin also hosts and API `/digests/{sandboxId}` that shows all the currently running container digests for a given sandboxId (or 404 if no such sandboxId). No containers
returns `{ "digests": [] }`
- hook this up to get-cert, so that get-cert fetches its sandboxId and requests its attestation with that
- on request to the CDS for a ratls cert, the sandboxId gets embedded in the cert (part of the signed area, so it's CDS-approved)
- the nri-image-policy signs the sandboxId on behalf of the get-cert request, bound to a CDS challenge nonce, so that:
  a) only a given get-cert can be calling the CDS.
  b) it's not replayable (the nonce is the single-use challenge for that issuance).
  c) CDS can prove the sandboxId came from the nri-image-policy.
- analogous changes in the in-kata policy-monitor
- renamed the 'workload-claims broker' to 'admission inventory'

## What Changed

The admission inventory (nri-image-policy on node-CVM, policy-monitor in the kata guest - the component previously called the workload-claims broker) now serves a sandbox-identity API behind one shared interface: 
- `POST /sandbox` issues an inventory-signed, nonce-fresh, caller-bound sandbox token
- `GET /digests/{sandboxId}` lists the image digests currently running in a sandbox (404 if unknown, `{"digests": []}` if empty). 

get-cert fetches the CDS challenge for the issuance, then a token bound to its CSR key and that challenge, and forwards it to CDS, which verifies it (EAR → attested inventory key → signature → nonce = this request's challenge → key binding) and stamps the sandbox ID into the signed leaf as extension `1.3.6.1.4.1.59888.1.4`.

## Why

A relying party holding a workload's RA-TLS leaf could not tell *which pod sandbox* stood behind it, and had no way to ask the node what is currently running there. This makes the NRI plugin the arbiter of sandbox identity: the leaf carries a CDS-approved sandbox ID, and `/digests/{sandboxId}` is the live cross-check keyed by it. 
That matters most during the claim-free first-issuance window (getcert-workload-binding.md, Corner 4): the sandbox ID binds at first issuance, so it is the leaf's only workload-identifying field until the workload digest binds.

The token construction:
- **Only the requesting get-cert can redeem it** - the token binds SHA-256 of the requester's public key; CDS requires it to match the CSR key, whose possession the CSR signature and evidence binding prove.
- **Fresh, not replayable, no clock** - the token carries the single-use CDS challenge for this issuance, and CDS requires it to equal the challenge it consumes for the request. The same nonce binds both the evidence REPORTDATA and the token, so the sandbox token rides the issuance's existing freshness instead of a wall-clock window - nothing to skew-tune, and a captured or pre-signed token can't be replayed against a later challenge.
- **CDS can prove inventory provenance** - the envelope carries a CDS-issued EAR (`POST /attest-key`) binding the inventory's signing key to TEE evidence on an allowed measurement; CDS validates it with its own JWKS, exactly as `/sign-csr` already does.

**Why no per-node cert/pk cache in CDS:** each envelope carries its own `/attest-key` EAR, so CDS verifies statelessly - no registration, no cache invalidation when an inventory restarts and mints a new key, and JWKS rotation is already handled by the rotator's overlap window. If we later want key-continuity pinning per node ("a second inventory key appeared for node X"), a cache becomes an observability feature, not a correctness one.

**Residual trust (documented in docs/ratls.md):** the EAR proves the signing key lives in a TEE on an allowed measurement. On node-CVM every process in the measured node shares that measurement, so "came from nri-image-policy" ultimately rests on the measured node image running only the intended components - the same honest-node assumption as getcert-workload-binding.md Corners 6–7. Hardening beyond that (e.g. an inventory-role claim gated on host-network source IP) is possible but out of scope here.

Other notes for review:
- The component is renamed "workload-claims broker" → **admission inventory** throughout (types, `pkg/workloadclaims` exports, docs, and the webhook-rendered get-cert flag `--workload-claims-broker` → `--workload-claims`); "broker" was overloaded with the incoming secret-broker. The flag rename needs no alias - it is only ever rendered by the webhook, which injects args and sidecar image together. The socket filename and routes are unchanged (they name the flow and are deployment-visible).
- `ObtainCertificateWithClaimsContext` gained `challenge string` and `sandboxToken json.RawMessage` parameters (breaking signature change; only in-repo caller is get-cert). The caller now fetches the challenge (via `AuthenticateContext`) so one nonce binds both the evidence and the token; the claims-free wrapper `ObtainCertificateWithEvidenceContext` fetches its own challenge, so mesh/other callers are unaffected. The token is opaque at the `pkg/types`/`pkg/attestclient` layer to avoid an import cycle (`types` ← `workloadclaims`).
- policy-monitor's inventory now records injected sidecars too and filters them at claim-query time (node-CVM parity), so the `/digests` inventory is complete; the workload digest is unchanged.
- Kata inventories learn the sandbox ID from the CRI `sandbox-id` annotations (pause container included) and only ever answer for their own single sandbox.

## Follow-up 
remove the extra claims from the workloadClaims, and use the sandboxID to infer them out of band instead


internal/cmds/getcert/run.go
```diff
@@ -323,24 +324,48 @@ func obtainCert(ctx context.Context, cfg config, client attestclient.Client) err
...
+// digest lists to forward to CDS, and the broker-signed sandbox token
+// (workloadclaims.SignedSandboxToken JSON) for CDS to verify and stamp. All
+// zero ⇒ the plain, claims-free flow.
 type workloadClaimsResult struct {
-	claimsDER   []byte
-	initDigests []string
-	mainDigests []string
+	claimsDER    []byte
+	initDigests  []string
+	mainDigests  []string
```
CluEleSsUK, line 357: I think we can actually drop these workload claims after adding the sandbox details, as we can use it to infer all the containers while running at better granularity

internal/cmds/nri-image-policy/plugin.go
```diff
@@ -205,6 +209,24 @@ func (p *plugin) RemoveContainer(ctx context.Context, pod *api.PodSandbox, ctr *
 	return nil
 }
 
+// RunPodSandbox records a started pod sandbox for the broker's sandbox set.
+// Only subscribed when the broker is enabled (see Configure).
+func (p *plugin) RunPodSandbox(ctx context.Context, pod *api.PodSandbox) error {
+	if p.broker != nil {
```
CluEleSsUK: should broker ever be nil here? if it is, is it worth erroring?

internal/cmds/nri-image-policy/plugin.go
```diff
@@ -205,6 +209,24 @@ func (p *plugin) RemoveContainer(ctx context.Context, pod *api.PodSandbox, ctr *
...
+		p.broker.recordSandbox(pod.GetId())
+	}
+	return nil
+}
+
+// RemovePodSandbox evicts a removed pod sandbox (and its containers) from the
+// broker. Only subscribed when the broker is enabled (see Configure).
+func (p *plugin) RemovePodSandbox(ctx context.Context, pod *api.PodSandbox) error {
+	if p.broker != nil {
```
CluEleSsUK: same here

internal/cmds/nri-image-policy/plugin.go
```diff
@@ -392,6 +414,14 @@ func (p *plugin) shouldCheckExisting() bool {
 func (p *plugin) Synchronize(ctx context.Context, pods []*api.PodSandbox, ctrs []*api.Container) ([]*api.ContainerUpdate, error) {
 	cfg := p.cfg
 
+	// Seed the broker's sandbox set now, even when the container check is
+	// deferred: sandbox existence needs no allowlist.
+	if p.broker != nil {
```
CluEleSsUK: and here

internal/cmds/policymonitor/monitor.go
```diff
@@ -420,6 +420,12 @@ func (m *monitor) handleNewContainer(ctx context.Context, dir string) {
 	// mislabelled workload can't slip through (kata would run the measured
 	// pause for it, not the host's image). Checked before extractDigest
 	// because the pause carries no image-name annotation.
+	// Every container (the pause included) names its pod sandbox in the CRI
+	// annotations; capture it for the broker's sandbox-identity surface.
+	if m.broker != nil {
```
CluEleSsUK: should this ever be nil?

internal/cmds/nri-image-policy/plugin.go
CluEleSsUK: Yes — nil is a real state: the broker only exists when `workload_claims.socket_dir` is configured; without it the plugin runs pure admission enforcement. For this hook (and RemovePodSandbox/RemoveContainer) nil additionally means the event was never subscribed (Configure only sets RUN/REMOVE_POD_SANDBOX with a broker), so the guard is defense against a runtime delivering an unsubscribed event. Erroring would turn that harmless no-op into a runtime-visible lifecycle-hook failure — strictly worse than ignoring. Documented the invariant once on the `broker` field rather than per-site.

internal/cmds/nri-image-policy/plugin.go
CluEleSsUK: Same answer as above — unsubscribed without a broker, so the guard only trips if the runtime misdelivers; no-op beats failing the teardown path. Invariant now documented on the `broker` field.

internal/cmds/nri-image-policy/plugin.go
CluEleSsUK: This one is different from the two hooks above: Synchronize is always invoked on plugin connect regardless of the event mask, so `broker == nil` here is the ordinary disabled-feature branch (no socket dir configured), not defense-in-depth — the guard is load-bearing and not an error case.

internal/cmds/policymonitor/monitor.go
CluEleSsUK: Yes: policy-monitor only constructs the broker when `--workload-claims-socket-dir` / `$C8S_WORKLOAD_CLAIMS_SOCKET_DIR` is set — a guest without the workload-claims mount runs monitor-only, and this callback still fires for every admitted container. So nil is ordinary configuration; erroring would kill enforcement over an optional feature being off. Noted on the `broker` field.

internal/cmds/getcert/run.go
```diff
@@ -133,9 +135,9 @@ alongside a workload that uses the obtained certificate.`,
...
-	flags.StringArrayVar(&cfg.WorkloadInitContainers, "workload-init-container", nil, "Name of a pod-spec init container (repeatable); the broker's containers are split into the init vs main image sets by these names for the workload digest (docs/ratls.md)")
+	flags.BoolVar(&cfg.WorkloadClaims, "workload-claims", false, "Bind a workload-digest claim by fetching this pod's admitted containers — and an inventory-signed sandbox token, which CDS verifies and stamps into the issued leaf — from the local inventory at get-cert's compiled Unix socket path (docs/ratls.md) — nri-image-policy on node-CVM, policy-monitor in the kata guest. The path is baked in, not supplied, so the control plane cannot redirect the fetch; fail-closed if the inventory is unreachable")
+	flags.DurationVar(&cfg.WorkloadClaimsTimeout, "workload-claims-timeout", 5*time.Second, "Timeout for the admission inventory fetch")
+	flags.StringArrayVar(&cfg.WorkloadInitContainers, "workload-init-container", nil, "Name of a pod-spec init container (repeatable); the inventory's containers are split into the init vs main image sets by these names for the workload digest (docs/ratls.md)")
```
AnomalRoil, line 140: honestly even after reading their details, I still have no idea how to use all three flags, or when to use them. Like when do I want to bind a workload-digest claim? What is such a claim anyhow? That's all quite abstract and not detailed in the docs afaict.

pkg/workloadclaims/sandboxtoken.go
```diff
@@ -0,0 +1,224 @@
...
+	ear, err := s.source(ctx, s.pubDER)
+	if err != nil {
+		return "", fmt.Errorf("workloadclaims: obtain inventory EAR: %w", err)
+	}
+	exp, err := jwtExpiry(ear)
+	if err != nil {
+		exp = time.Time{}
+	}
+	s.ear, s.earExp = ear, exp
+	return ear, nil
```
AnomalRoil, line 209: should we really accept to sign a token with no expiry date??

AnomalRoil (approved):
How does one obtain the public key of a `SandboxTokenSigner` ??

LGTM globally. Let's merge 161 onto this and do a last pass

pkg/workloadclaims/sandboxtoken.go
CluEleSsUK: it is bounded by the challenge nonce from the CDS anyway

internal/cmds/getcert/run.go
CluEleSsUK: basically automated with the injection based on charts - prob don't want to run it by hand

CluEleSsUK:
>How does one obtain the public key of a SandboxTokenSigner ??

you don't, but the CDS verifies it and issues an EAR to the NRI/policy-monitor create on start. CDS reads it from the EAR

PR #152, opened 2026-07-27
ci: add advisory diff-based mutation testing

joaosa:
## What

Adds an advisory mutation-testing job that grades PR diffs with gremlins v0.6.0, pinned as a go.mod tool dependency.

- `scripts/mutation-check.sh`: `run` (diff-based gremlins run, writes json + log), `summary` (markdown report), `ci` (run + summary into the step summary, always exits 0).
- `.gremlins.yaml`: excludes `internal/helmchart` (tests exec a real helm binary) and generated deepcopy code.
- New PR-only **Mutation (advisory)** CI job: mutates only code changed vs the base ref and posts efficacy plus surviving mutants to the job step summary. `continue-on-error` and exit-0 semantics mean it can never block a PR.
- `make mutation-check` for local runs (`BASE=<ref>` picks the base, default origin/main).

Line coverage says whether code ran under a test; mutation score says whether any test would notice the code being wrong. The coverage ratchet from #137 can be satisfied by assertion-light tests, so this job is the instrument that tells those apart from tests that actually pin behavior.

## Verification
- flag/config/report-field names pinned against a real gremlins v0.6.0 run (`--diff`, `--exclude-files`, json field and status names)
- linux run over the allowlist feature diff (vs 3ace878): 269 mutants graded in about 3.2 minutes of gremlins time, comfortably inside the 20 minute job cap
- 0-mutant, missing-report, and tool-crash paths all exit 0 and render a note in the summary
- the first run's findings and the tests that kill them are the stacked follow-up PR


CluEleSsUK:
lgtm! the github suggests it doesn't work well for big modules, but a 2m run isn't bad

CluEleSsUK (approved):

PR #155, opened 2026-07-27
test: cover controller Run, cds handoff snapshot, and ratlsmesh clientset wiring

joaosa:
## What

Three wiring paths had no tests because nothing could reach them: controller `Run` calls `ctrl.GetConfigOrDie`/`ctrl.NewManager`/`discovery.NewDiscoveryClientForConfig` directly, the cds handoff `Snapshot` closure is only invocable through a full attested /handoff exchange, and ratlsmesh's clientset construction requires a real in-cluster environment.

Behavior-preserving seams in the pattern the repo already uses (`newDirectClient`, `newKubeClientset` package vars):

- `internal/controller`: the three constructors become package vars; `Run`'s body is otherwise unchanged.
- `internal/cmds/cds`: the `Snapshot` closure body moves verbatim into `snapshotAllowlist(load, mesh)`; the closure now just calls it.
- `internal/cmds/ratlsmesh`: `rest.InClusterConfig` hoisted to an `inClusterConfig` var behind the existing `newKubeClientset` seam.

New tests cover the previously unreachable paths: manager-create, discovery-client, and manager-exit errors, plus the invariant that disabling the status mirror skips discovery entirely; the handoff snapshot's success shape and its two withholding paths (load failure, unparsable stored digest); both clientset outcomes off-cluster.

Also adds `scripts/mutation-check.sh full` and `make mutation-full` for whole-module runs (the CI job stays diff-based), lists `make mutation-check` in the README dev commands, and adds a `selftest` subcommand run from Build & Test: the advisory job only exercises the 0-mutant path on script-only diffs, so the summary rendering paths (survivors, zero mutants, missing report) are checked against fixtures in `scripts/testdata/`.

## Verification
- `go test -race -count=1` green on the three packages (darwin, plus a linux container for the linux-tagged ratlsmesh tests)
- every new error-path test checked by inverting its guarded condition and confirming the test fails (6 sites, ratlsmesh on linux)


CluEleSsUK (changes requested):
we should make sure that the mutation-check runs whenever the shell script for it is modified

joaosa:
> we should make sure that the mutation-check runs whenever the shell script for it is modified

the ci paths includes scripts/mutation-check.sh and .gremlins.yaml, so script changes trigger the full workflow including the advisory job. 

a script-only diff would exercise the 0-mutant path though, so this pr adds mutation-check.sh selftest plus fixtures.

AnomalRoil (approved):
lgtm

scripts/testdata/summary-sample.json
CluEleSsUK, line 1: should this be committed?

scripts/testdata/summary-zero.json
CluEleSsUK, line 1: Also this

CluEleSsUK (approved):

scripts/testdata/summary-zero.json
joaosa: moved these into the test report generation

PR #156, opened 2026-07-27
feat(ci): run the TDX bare-metal lane on merge alongside SNP

b0xtch:
Makes a merge to main exercise **TDX bare metal** as well as SEV-SNP. Depends on confidential-ci PR (adds `workflow_call` to the source lane).

## Why a sibling job, not a matrix row

A reusable-workflow `uses:` must be a static literal and cannot be driven by `${{ matrix }}` (confidential-ci `research/ADR-matrix-architecture-2026-07-20.md`, line 29). So TDX cannot be `- { platform: tdx-metal, ... }` in the existing matrix.

It would want to be a separate job regardless: the two lanes use genuinely different transports, console for SNP and RA-TLS for TDX. Folding them into one reusable workflow is exactly the two-drivers-in-one-file shape that was just removed from the upstream `cvm-e2e.yml`.

It carries no `needs: payload`, since the TDX lane is self-contained and boots on a different host, so the two lanes run in parallel rather than one waiting on the other.

## The lane is proven, not aspirational

confidential-ci run [30238209289](https://github.com/confidential-dot-ai/confidential-ci/actions/runs/30238209289): **all 17 steps green** — boot, genuine-TDX assertion (`qemu -object tdx-guest`), attestation-api reporting `"platform":"tdx"`, RA-TLS operator-key-bound kubeconfig, `c8s install`, converge, the **negative** floor-denies-unlisted-image assertion, and teardown.

That negative test is the valuable one: it proves the product's fail-closed admission floor actually denies an unlisted image, on real TDX hardware, per run.

## Vendoring and drift, which you asked about

c8s is public and confidential-ci is private, so a public repo cannot `uses:` the private reusable workflow. A copy is unavoidable.

To stop the copies silently separating:

- **Both copies of `tdx-metal-e2e.yml` carry both triggers** (`workflow_call` + `workflow_dispatch`), so no per-repo edit is ever needed and a plain checksum can prove they match.
- **A provenance banner** names the source of truth, the vendored copy, and the rule: edit upstream, then copy the whole file across.

This PR also puts banners on the **two files that were already vendored with no pointer home at all**: `cvm-e2e.yml` and `e2e-c8s.yml`. That absence is precisely why this repo's `cvm-e2e.yml` fell **121 lines** behind its upstream and lost an entire platform arm with nothing detecting it.

I did **not** force those two back into sync. They have diverged over real time and reconciling them is separate, deliberate work; the banners exist so the next person knows to look rather than assuming the files are the same.

## Verification

YAML parses for all three files, every `run:` block passes `bash -n`, and the caller passes only inputs the callee declares (`c8s_ref`).

**Not yet exercised end to end from this repo.** A PR cannot trigger the merge path, and `runs-on: cvm-launcher-tdx` is org self-hosted hardware which must never be exposed to `pull_request`. Worth dispatching `e2e-c8s.yml` on this branch before merging.

## Interaction with the trigger fix

With the current path gate, this new job inherits the same hole: a merge touching no source paths fires neither lane. #143 fixes that at the head of the chain.

### DIFF
--- a/.github/workflows/cvm-e2e.yml
+++ b/.github/workflows/cvm-e2e.yml
@@ -1,4 +1,21 @@
 # ═══════════════════════════════════════════════════════════════════════════
+# VENDORED FILE. SOURCE OF TRUTH LIVES IN confidential-ci.
+# ═══════════════════════════════════════════════════════════════════════════
+# UPSTREAM: confidential-dot-ai/confidential-ci  .github/workflows/cvm-e2e.yml
+# THIS FILE: confidential-dot-ai/c8s             .github/workflows/cvm-e2e.yml
+#
+# WHY A COPY: c8s is PUBLIC and confidential-ci is PRIVATE, and a public repo
+# cannot `uses:` a private repo's reusable workflow. Vendoring is the stopgap.
+#
+# THESE TWO COPIES HAVE ALREADY DIVERGED, and it went unnoticed: upstream grew
+# a tdx-metal arm this copy never had, then had it removed again. Neither file
+# referenced the other, so nothing could detect it. Reconciling them is separate
+# work; this banner exists so the next person knows to look.
+#
+# WHEN CHANGING THIS FILE: change the upstream copy too, or say in the PR why the
+# two must differ. A silent one-sided edit is how the drift above happened.
+#
+# ═══════════════════════════════════════════════════════════════════════════
 # THE PRIMITIVE — "run CI inside a CVM", reusable across the whole stack.
 # ═══════════════════════════════════════════════════════════════════════════
 #
--- a/.github/workflows/e2e-c8s.yml
+++ b/.github/workflows/e2e-c8s.yml
@@ -1,4 +1,20 @@
 # ═══════════════════════════════════════════════════════════════════════════
+# VENDORED FILE. SOURCE OF TRUTH LIVES IN confidential-ci.
+# ═══════════════════════════════════════════════════════════════════════════
+# UPSTREAM: confidential-dot-ai/confidential-ci  .github/workflows/e2e-c8s.yml
+# THIS FILE: confidential-dot-ai/c8s             .github/workflows/e2e-c8s.yml
+#
+# WHY A COPY: c8s is PUBLIC and confidential-ci is PRIVATE, and a public repo
+# cannot `uses:` a private repo's reusable workflow. Vendoring is the stopgap.
+#
+# This copy and its upstream have diverged over time. Treat upstream as the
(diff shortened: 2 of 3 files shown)


b0xtch:
Validated end to end from this branch: [run 30288869037](https://github.com/confidential-dot-ai/c8s/actions/runs/30288869037), **both lanes green from one trigger**.

```
success  e2e / payload
success  e2e / tdx / c8s on TDX metal
success  e2e / e2e (snp-metal, cvm-launcher) / cvm (snp-metal/rke2-node)
```

The TDX job did the real work, not just resolve the `uses:` — all 17 steps green, including the ones that matter:

```
success  boot the TDX CVM
success  assert genuine TDX launch
success  get an attested, operator-key-bound kubeconfig (RA-TLS)
success  install the c8s bundle under test
success  assert components converged
success  NEGATIVE, the baked floor must deny a non-allowlisted image
success  teardown
```

The two lanes ran **in parallel on different hosts**, which is why there is no `needs: payload` on the TDX job.

Since this run was dispatched, two follow-up commits landed on the branch:

1. **The paired-ref fix.** The job originally passed `c8s_ref: ${{ inputs.c8s_ref }}`, which on the merge path is the merge head. The TDX node image bakes its fail-closed floor from its own build ref (`64b6d7a`, while c8s main is at `6522581`), so an unpaired ref gets its own components denied and TDX would have gone red on every merge. The job now passes no ref and the constraint is documented in `e2e-c8s.yml` so it does not get "fixed" back.

2. **Naming consistency**, so the two lanes scan as a pair rather than unrelated things:
   ```
   before:  e2e / tdx / c8s on TDX metal
   after:   e2e / tdx (tdx-metal, cvm-launcher-tdx) / cvm (tdx-metal/rke2-node)
   ```

Both are cosmetic-or-safer relative to what was validated: the rename cannot change behaviour, and the ref change makes it strictly less likely to fail. Happy to re-dispatch if you would rather the merged head be the exact tree that ran.

**Merge order:** confidential-ci#46 first (source of truth for the rename), then this. The two copies of `tdx-metal-e2e.yml` verify byte-identical at `61a7273377b522c4…`.

b0xtch:
Superseded by confidential-ci#48, now merged.

This wired TDX as a sibling job because I believed a matrix row was impossible. The `uses:` restriction is real, but the conclusion was not: all four platforms are now first-class jobs in confidential-ci's `e2e-c8s.yml`, which is down to 111 lines and contains nothing but the platform list.

TDX will reach this repo through the vendoring stage instead, alongside the other three lanes and on the converged shape, rather than as a one-off sibling job that would have needed reworking immediately.

PR #158, opened 2026-07-27
ci: detect drift between vendored workflows and their source of truth

b0xtch:
c8s vendors copies of confidential-ci's reusable workflows because a public repo cannot `uses:` a private one. **Nothing has ever checked that the copies still match.** That is exactly how `cvm-e2e.yml` here drifted **121 lines** behind its original and lost an entire platform arm with nobody noticing.

Compares the sha256 of each vendored workflow against its source of truth.

## Two tiers, so it is adoptable today

| tier | files | behaviour |
|---|---|---|
| **blocking** | `tdx-metal-e2e.yml` | must be byte-identical; it carries both triggers in both repos precisely so that is achievable |
| **advisory** | `cvm-e2e.yml`, `e2e-c8s.yml` | already diverged (422 vs 450, 299 vs 337); reported so the gap stops growing |

Reconciling the historical drift is separate work. Gating on it today would just block every PR touching those files, so this reports the delta instead and blocks only where byte-identity is actually the contract.

Runs on PRs touching those files, **and weekly**, because drift introduced from the *upstream* side would never appear in a c8s PR at all.

## It needs a secret, and it fails loudly without one

`CONFIDENTIAL_CI_READ_TOKEN`, read-only, scoped to the private upstream. Until it exists the check reports **SKIPPED with a warning** rather than passing, so a missing secret can never masquerade as a clean result.

Adding that secret is a permission change, so per the org process it needs an Access Change Request rather than an ad-hoc grant. I have drafted one; see the comment below.

Runs on `ubuntu-latest`, never self-hosted, so the `pull_request` trigger carries no confidential-hardware exposure. Fork PRs receive no secrets and will correctly report SKIPPED.

### DIFF
--- a/.github/workflows/vendor-drift-check.yml
+++ b/.github/workflows/vendor-drift-check.yml
@@ -1,0 +1,93 @@
+# ═══════════════════════════════════════════════════════════════════════════
+# VENDOR DRIFT CHECK
+# ═══════════════════════════════════════════════════════════════════════════
+# c8s is PUBLIC and confidential-ci is PRIVATE, so a public repo cannot `uses:`
+# a private repo's reusable workflow and c8s carries vendored copies instead.
+# Vendoring without a check is how you get silent divergence: c8s's cvm-e2e.yml
+# drifted 121 lines behind its original and lost an entire platform arm, and
+# nothing detected it because neither copy referenced the other.
+#
+# This compares the sha256 of each vendored file against its source of truth.
+# It is deliberately advisory for files KNOWN to have already diverged, and
+# blocking for files that are supposed to be byte-identical, so it can be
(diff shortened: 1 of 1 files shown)


b0xtch:
Closing as premature. Reopening this at the point it has something to guard.

The blocking tier here protects `tdx-metal-e2e.yml` byte-identity, but that file is not vendored into this repo yet, so there is nothing for it to block on. The advisory tier reports drift that is already known and measured. Meanwhile it needs a credential provisioned to do anything at all, so as things stand it costs a permission request and returns a SKIPPED status.

Better to add it alongside the vendoring change itself, when the set of files that must stay identical is settled rather than assumed.

PR #159, opened 2026-07-27
feat(ci): run both metal lanes on merge, and reconcile the vendored drift

b0xtch:
A merge to c8s main runs **one** lane today. The four-lane work landed in confidential-ci's copy, and this repo runs its own vendored copy, which was never updated. This brings it across.

## Drift reconciled first, not buried

`cvm-e2e.yml` here was **422 lines against 450** upstream. I checked what that gap actually was before overwriting anything:

**This copy already had every safety fix** — the readonly ROX rootdisk, the configfs-tsm measurement read, the cross-repo reaper, the pinned igvm sidecar digest. The gap was upstream growing standing-runner features (`keep_alive`, `vm_name`, `memory_override`) that this repo never used.

All three are optional and default inert, and:

```yaml
VM: ${{ inputs.vm_name || format('cvm-{0}-{1}', github.run_id, github.run_attempt) }}
```

produces **exactly** the VM name this copy already produced when `vm_name` is unset. So the sync is behaviour-preserving, not just plausible.

The three vendored lanes are now **byte-identical** to their originals:

| file | sha256 |
|---|---|
| `cvm-e2e.yml` | `7439704ebad6` |
| `snp-metal-e2e.yml` | `dc0629dae1f4` |
| `tdx-metal-e2e.yml` | `85bae38ea62c` |

## Shape

`e2e-c8s.yml` goes 300 lines to 98 and becomes only the platform list. The SNP payload moves into `snp-metal-e2e.yml`, because `cvm-e2e.yml` is a **generic primitive** shared with other callers and the c8s payload does not belong inside it.

Not a `matrix:`: a reusable-workflow `uses:` takes no expressions, so a matrix needs an if-gated dispatcher, and since every row instantiates the whole dispatcher that costs `rows x jobs` tree entries, growing quadratically. Plain jobs give one real entry each.

## The azure rows are absent on purpose

This is the **one** intended difference from the upstream copy, and both blockers are outside CI:

1. The Entra federated credential trusts `repo:confidential-dot-ai/confidential-ci`. A token minted from this repo is rejected with `AADSTS700213`. **Observed, not assumed**: the same rejection happened on a confidential-ci feature branch, and succeeded once on main. Granting c8s its own credential is an access change, not a code change.
2. `azure-tdx-e2e.yml` reads `c8s-e2e/azure-tdx-driver.sh`, which lives in confidential-ci and returns 404 here.

Until both are resolved the azure lanes keep running from confidential-ci on a schedule, so coverage is unchanged, just not merge-triggered from here.

## Verification

Validated by dispatch on this branch, since a PR cannot exercise a self-hosted lane. Result in a comment below.

The gate is not a green check: the failure mode this system has hit before is a run that stays green while measurement enforcement silently degrades to `MEAS_ACCEPTANY`. The check is that the emitted digest still equals the recorded baseline `15bc9953...d19115`, which independently matches the cluster's own `rke2-image-refs.runtimeLaunchDigest`.

## Note on what this does and does not fix

This makes both metal lanes run **when the merge trigger fires**. It does not fix the trigger. `docker.yml`'s push path gate still means roughly a quarter of merges to main produce no Docker run and therefore no TEE integration at all. That is #143.

### DIFF
--- a/.github/workflows/cvm-e2e.yml
+++ b/.github/workflows/cvm-e2e.yml
@@ -27,7 +27,7 @@ on:
   workflow_call:
     inputs:
       platform:
-        description: 'target cell: snp-metal | tdx-metal | gke-snp | gke-tdx'
+        description: 'target cell: snp-metal (the only wired cell). TDX metal is NOT here: it has its own console-free lane, tdx-metal-e2e.yml. gke-snp and gke-tdx are unwired stubs.'
         type: string
         required: true
       runner:
@@ -45,6 +45,18 @@ on:
       timeout_minutes:
         type: number
         default: 55
+      keep_alive:
+        description: 'leave the CVM running after PAYLOAD_OK (standing-runner provisioner) — skips teardown; the VM is labelled standing so the reaper spares it'
+        type: boolean
+        default: false
+      vm_name:
+        description: 'stable VM name for a standing CVM (default: cvm-<run_id>-<attempt>, ephemeral)'
+        type: string
+        default: ''
+      memory_override:
+        description: 'override the per-platform guest RAM (e.g. 48Gi for a standing multi-runner CVM). vCPU is NOT overridable — it is baked into the measured IGVM (smp4). RAM is not in the launch measurement, so this keeps the same golden digest. Empty = per-platform default.'
+        type: string
+        default: ''
     secrets:
       ghcr_user:
         required: false
@@ -65,7 +77,7 @@ jobs:
       measurement: ${{ steps.run.outputs.measurement }}
     env:
       NS: confai-images
-      VM: cvm-${{ github.run_id }}-${{ github.run_attempt }}
+      VM: ${{ inputs.vm_name || format('cvm-{0}-{1}', github.run_id, github.run_attempt) }}
     steps:
       - name: platform config
         run: |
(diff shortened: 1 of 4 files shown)


b0xtch:
Validated by dispatch on this branch: [run 30302649119](https://github.com/confidential-dot-ai/c8s/actions/runs/30302649119), **both metal lanes green from this repo**.

```
success  e2e / snp-metal / payload
success  e2e / snp-metal / cvm / cvm (snp-metal/rke2-node)
success  e2e / tdx-metal / cvm (tdx-metal/rke2-node)
```

**The gate passes from c8s**, which is the point of running it here rather than trusting the upstream result:

```
digest    15bc9953...d19115   ==  BASELINE_MEAS
MEAS_ACCEPTANY  absent
```

So the vendored `cvm-e2e.yml` is proven equivalent, not assumed to be. That matters because the drift being reconciled was 28 lines, and a silent measurement regression is the one failure mode here that still reports green.

For the record, the three vendored lanes verify byte-identical to their confidential-ci originals:

| file | sha256 |
|---|---|
| `cvm-e2e.yml` | `7439704ebad6` |
| `snp-metal-e2e.yml` | `dc0629dae1f4` |
| `tdx-metal-e2e.yml` | `85bae38ea62c` |

Net effect once merged: a merge to main runs **two** TEE lanes instead of one, on real SEV-SNP and real Intel TDX hardware.

Still not fixed by this PR, and worth stating plainly: `docker.yml`'s push path gate means roughly a quarter of merges produce no Docker run and therefore trigger none of this. That is #143.

PR #161, opened 2026-07-28
fix(workloadclaims): close SO_PEERCRED PID-reuse TOCTOU

CluEleSsUK:
SO_PEERCRED returns a bare PID, and the inventory reads /proc/<pid>/cgroup a few instructions (and one RLock) later. In that window the peer can exit and the kernel recycle its PID to a process in another container, which the inventory would then resolve as the caller — binding a get-cert leaf's sandbox ID / workload digests to a victim's pod.

Read SO_PEERPIDFD (Linux 6.5+) alongside SO_PEERCRED: a pidfd pins the exact process instance. The resolver rechecks peer.Valid() AFTER the /proc read; if the pinned process exited during resolution the answer is refused (fail closed). Pre-6.5 kernels have no pidfd and degrade to the prior best-effort behavior rather than failing every request.

- New Peer handle (peercred_linux.go / peercred_other.go stub) carrying the PID plus a pidfd, with PID()/Valid()/Close(). Resolver.ContainersForPeer and SandboxResolver.SandboxForPeer now take a Peer; the kata resolver ignores it (guest boundary is the isolation), the node-CVM resolver rechecks liveness in callerForPeer.
- Orthogonal to the shallowest-tracked-cgroup defense (Corner 2): pidfd fixes PID identity over time, that rule fixes cgroup-nesting spoofing.
- Test spawns a child, pins it, kills it, and asserts Valid() flips to false — proving the recheck rejects a recycled PID.

Pre-existing weakness on the /v1/workload-digests path; the sandbox routes inherit the same fix. See docs/getcert-workload-binding.md, Corner 1 ("Pinning the PID against reuse").

### DIFF
--- a/pkg/workloadclaims/peercred_linux.go
+++ b/pkg/workloadclaims/peercred_linux.go
@@ -1,31 +1,113 @@
 package workloadclaims
 
 import (
+	"errors"
 	"net"
 
 	"golang.org/x/sys/unix"
 )
 
-// peerPID returns the kernel-reported PID of the process on the other end of
-// a unix socket, or 0 when unavailable (non-unix transport, or a failed
-// getsockopt). The node-CVM resolver must reject 0 — it binds the caller to a
-// pod; the kata resolver ignores the PID entirely, one pod per guest
-// (docs/ratls.md).
-func peerPID(c net.Conn) int {
+// pidfd sentinels for Peer.pidfd. A real pidfd is >= 0.
+const (
+	// pidfdNone: peerFrom obtained no pidfd (non-unix conn, or SO_PEERPIDFD
+	// failed). A CC node always supports SO_PEERPIDFD — it landed in Linux 6.5,
+	// and SNP hosts require >= 6.11, TDX >= 6.16 — so its absence is anomalous,
+	// not an old-kernel we tolerate. IsAlive fails closed on it.
+	pidfdNone = -1
+	// pidfdSkip: a test-constructed Peer (PeerForPID) with no socket, so there
+	// is no liveness to check. IsAlive returns true so resolver tests can run
+	// without a real process. Never produced by peerFrom.
+	pidfdSkip = -2
+)
+
+// Peer identifies the process on the other end of an inventory connection. The
+// PID comes from SO_PEERCRED; SO_PEERPIDFD (Linux 6.5+) additionally returns a
+// pidfd that pins that exact process instance, so IsAlive can detect an exit
+// between the credential read and the resolver's /proc lookup — the window a
+// bare PID leaves open to PID reuse (docs/getcert-workload-binding.md,
+// "Corner 1"). Close releases the pidfd.
+type Peer struct {
+	pid   int
+	pidfd int // >= 0: a live pidfd to poll; pidfdNone/pidfdSkip otherwise
+}
+
+// peerFrom captures the peer credentials of a unix connection. A non-unix conn
+// or a failed SO_PEERCRED yields a zero-PID Peer, which the node-CVM resolver
+// rejects. SO_PEERPIDFD must succeed on a CC node; any failure leaves pidfdNone,
+// and IsAlive then fails closed rather than vouching for a caller it cannot pin.
+func peerFrom(c net.Conn) Peer {
+	p := Peer{pidfd: pidfdNone}
 	uc, ok := c.(*net.UnixConn)
 	if !ok {
-		return 0
+		return p
 	}
 	raw, err := uc.SyscallConn()
 	if err != nil {
-		return 0
+		return p
 	}
-	var cred *unix.Ucred
-	var credErr error
-	if err := raw.Control(func(fd uintptr) {
-		cred, credErr = unix.GetsockoptUcred(int(fd), unix.SOL_SOCKET, unix.SO_PEERCRED)
-	}); err != nil || credErr != nil || cred == nil {
-		return 0
-	}
-	return int(cred.Pid)
+	_ = raw.Control(func(fd uintptr) {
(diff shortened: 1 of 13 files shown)


pkg/workloadclaims/peercred_other.go
```diff
@@ -4,10 +4,25 @@ package workloadclaims
...
+
+// Valid is best-effort true; PID 0 is rejected upstream regardless.
+func (Peer) Valid() bool { return true }
+
+// Close is a no-op off Linux.
+func (Peer) Close() {}
+
+// PeerForPID builds a Peer from an out-of-band PID (tests). See the Linux
+// implementation for the production path.
+func PeerForPID(pid int) Peer { return Peer{pid: pid} }
```
AnomalRoil, line 29: Should this really instantiate the `Peer` with `pidfd==0` or should it be `-1` here?

pkg/workloadclaims/workloadclaims.go
```diff
@@ -291,6 +293,13 @@ type SandboxResolver interface {
...
+// peerFromRequest pins the caller of an inventory request. The returned Peer's
+// pidfd must be released with Close once resolution is done.
+func peerFromRequest(r *http.Request) Peer {
+	conn, _ := r.Context().Value(connKey{}).(net.Conn)
+	return peerFrom(conn)
+}
+
 // maxSandboxRequestBytes bounds a SandboxPath request body; it carries one
 // PKIX public key.
 const maxSandboxRequestBytes = 64 << 10
```
AnomalRoil, line 305: totally outside of that review, but what kind of magic number definition is that? Either `1 << 16` or `8 << 13` if the idea is that there's a 13 char limit on the bytes or whatnot, but that's just totally random as is, no ?

pkg/workloadclaims/peercred_linux.go
```diff
@@ -6,26 +6,83 @@ import (
...
+func (p Peer) PID() int { return p.pid }
+
+// Valid reports whether the pinned peer process is still alive. It MUST be
+// called AFTER the resolver reads /proc for this PID: a peer that exited during
+// resolution may have had its PID recycled, so the read cannot be trusted and
+// the caller must fail closed. With no pidfd (kernel < 6.5) it returns true —
+// best-effort, matching the pre-pidfd behavior.
+func (p Peer) Valid() bool {
+	if p.pidfd < 0 {
+		return true
```
AnomalRoil: This here seems very very weird, since right now most constructors for `Peer` are defaulting `pidfd` to `-1`. I'm pretty sure this means we wouldn't catch a regression or a `pidfd` that's never set instead of it being properly set.

pkg/workloadclaims/peercred_linux.go
```diff
@@ -6,26 +6,83 @@ import (
...
+		cred, err := unix.GetsockoptUcred(int(fd), unix.SOL_SOCKET, unix.SO_PEERCRED)
+		if err != nil || cred == nil {
+			return
+		}
+		p.pid = int(cred.Pid)
+		// SO_PEERPIDFD reflects the peer captured at connect(), consistent with
+		// SO_PEERCRED above, so the pidfd pins the same process the PID names.
+		if pidfd, err := unix.GetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_PEERPIDFD); err == nil {
+			p.pidfd = pidfd
+		}
```
AnomalRoil, line 60: On a 6.4-or-older kernel your `GetsockoptInt` returns `ENOPROTOOPT` and you silently leave `p.pidfd` at `-1`, I guess this is your goal? might be explicited

pkg/workloadclaims/peercred_linux.go
```diff
@@ -6,26 +6,83 @@ import (
...
+	for {
+		_, err := unix.Poll(fds, 0)
+		if err == unix.EINTR {
+			continue
+		}
+		if err != nil {
+			return false // cannot confirm liveness → fail closed
+		}
+		return fds[0].Revents&unix.POLLIN == 0
 	}
```
AnomalRoil, line 98: First, we should avoid infinite loops without `Context` cancelation, imo, or at least a timeout: that's a risk of DOS / leaking go routines.

Next, `Valid` is a terrible name for `IsAlive`... 

Finally, what if your `fd` is bad? You might want a little:
```
		// POLLNVAL/POLLERR/POLLHUP: fd is bad or hung up
		if fds[0].Revents&(unix.POLLNVAL|unix.POLLERR|unix.POLLHUP) != 0 {
			return false
		}
```

Last but not least, any reasons not to rely on `os.FindProcess` and the likes? Since Go 1.23, these are using `pidfd` under the hood on supported platform: https://go.dev/doc/go1.23#ospkgos

pkg/workloadclaims/peercred_linux.go
AnomalRoil: I think this should fail open only on `ENOPROTOOPT` (genuinely old kernel), btw, so that another failure wouldn't be ignored.

AnomalRoil (changes requested):

pkg/workloadclaims/peercred_linux.go
CluEleSsUK: it's not about regression but rather supporting old kernels; kernels below 4.6 won't don't have SO_PEERPIDFD, only SO_PEERCRED so this will return -1

pkg/workloadclaims/peercred_linux.go
CluEleSsUK: fair, we could reject old kernels - I don't think we've decided how old we are willing to support, but I think CC generally needs a newer kernel so prob safe

PR #162, opened 2026-07-28
test: strengthen weak tests across the security core and cds command packages

joaosa:
## What

Across sixteen packages, existing tests exercised code without pinning its behavior: assertion-light paths in the CA/handoff/CSR stack, RA-TLS verification, the attestation clients, and the cds/verify/getcert/credrelease commands would keep passing under real bugs (inverted validation guards, off-by-one boundaries, dropped error checks). This adds ~360 targeted tests that pin those behaviors; no production code changes.

Highlights of what is now pinned:
- `internal/issuer`: JWKS rotation lookups, handoff retry/refresh flows, CA rotator bundle publication, rate-limit keying by source IP, CSR policy SAN edge cases.
- `pkg/ratls`: attested-cert extension parsing, verifier config-claim pins, cert rotation deadlines, and the invariant that a server or client configured without a CA cert never gains a CA-trust path.
- `internal/cmds/cds`: RA-TLS warm-up fails closed, /attest-key operator-policy gate returns 403 on hash mismatch vs 400 on bad challenge, key-rotation loop stays off at interval zero, allowlist writes honor JWT clock skew.
- `internal/cmds/verify` + `internal/localverify`: inline VCEK verifies offline (never contacts KDS), --timeout bounds collateral fetches, TCB/text rendering sections.
- `pkg/workloadclaims`: unix-socket ownership/mode invariants (verified on linux, incl. root-only branches), digest verification role swaps.
- getcert renewal loop reload/warn paths, credrelease token clock skew and cert validity window, rtmr3measurer scan throttling and record ordering, probefile interval clamping.

Also excludes `test/` mock fixtures from mutation runs (`.gremlins.yaml`): their mutants grade the fixtures, not the product.

Measured with `make mutation-full`: efficacy 87.3% to 92.9%, surviving mutants 369 to 208 (2915 gradable). What remains lives in ratlsmesh, webhook, controller, policymonitor, nri-image-policy, and cmd/c8s, which are the next round.

## Verification
- `go test -race -count=1 ./...` green on linux (full suite in a container); darwin green for all non-linux-tagged packages
- ~75 of the new tests checked by inverting the guarded condition and confirming the test fails (linux-only ones inside the container)
- equivalent or seam-requiring survivors documented per package during review of the run output


AmeanAsad:
@joaosa sanity checking have we pruned this of string checking and no-logic tests?

joaosa:
> @joaosa sanity checking have we pruned this of string checking and no-logic tests?

yes I have just reviewed it

PR #165, opened 2026-07-28
test: cover the install, uninstall, and tee-label CLI flows

joaosa:
## What

`c8s install` and `c8s uninstall` are the install vector and had no tests covering their orchestration: preflights, helm/kubectl/crane sequencing, values construction, the kata host sweep, TEE node labelling, and render-values. This adds ~2,000 lines of behavior-pinning tests; no production code changes.

The tests drive the real cobra command tree against stub helm/kubectl/crane binaries on PATH (each invocation logged), asserting:
- exact argv sequences and ordering (preflights before namespace apply before `helm upgrade`; adoption patch after `--wait`)
- the computed values helm receives, decoded as YAML (cvmMode, teeDevices, digest-vs-tag resolution, component toggles, derived tls-lb upstream)
- the image-tag scheme end to end (unstamped builds resolve `:main`; explicit `--image-tag` verbatim; release tags untouched)
- uninstall's kata host sweep (helm-uninstall / apply / rollout / delete ordering, drain-wait selectors, typed decode of the sweep DaemonSet, running-pod refusal vs `--force`, unsafe-path refusals, `--delete-crds`/`--delete-namespace`)
- TEE auto-labelling per platform incl. conflict refusal, and render-values bundle output
- fail-fast paths proven by the absence of later argv in the call log

## Verification
- `go test -race -count=1 ./cmd/c8s/` green (darwin and linux container)
- 14 of the pinned behaviors checked by inverting the guarded condition and confirming the test fails
- whole-module mutation coverage rises from 90.5% to 95.1% (not-covered mutants 318 to 165) with efficacy steady at 97.5%


AmeanAsad (approved):
:shipit:

PR #166, opened 2026-07-28
test(helmchart): assert on decoded manifests instead of rendered text

joaosa:
## What

chart_test.go pinned rendered output with substring matches against the raw YAML stream (60 `strings.Contains` plus exact-text checks), so key reordering, indentation, or a rename in an intermediate value broke tests with zero behavior change. Every raw-stream assertion now decodes the rendered manifests into typed structs and asserts on fields: container argv and env, RBAC verb tables per resource, parsed nginx directives and server blocks, VAP CEL expressions, ConfigMap data entries, pod-template checksum annotations, and a recursive decoded-tree sweep for pull-secret refs.

Deliberately kept as strings: `helm template` failure messages (user-visible contracts), Contains against already-decoded field values (script bodies, CEL text, the quoted CORS header list), and the tee-proxy-remnant name sweep (a pure name search cannot false-positive on formatting).

New helpers extend the file's existing typed patterns: `renderedWorkload`/`renderedPodSpecs`, `renderedManifestHasLabel`, `kataEnforcementExpressions`, `findKey`, and nginx server-block parsing with `assertNoDirectivePrefix`.

## Verification
- `go test -count=1 ./internal/helmchart/` green, also with `-race`
- 16 assertions checked by mutating the corresponding chart template (image, flag, verb, annotation, secret ref), confirming the typed test fails, then reverting; helmchart is excluded from mutation runs so this manual pass stands in for the gremlins check
- chart templates untouched: the diff is one test file

### DIFF
--- a/internal/helmchart/chart_test.go
+++ b/internal/helmchart/chart_test.go
@@ -164,17 +164,17 @@ func TestChartDefaultRendersReplacementStack(t *testing.T) {
 	if !renderedManifestHasKind(t, out, "MutatingWebhookConfiguration") {
 		t.Fatalf("default chart missing MutatingWebhookConfiguration\n%s", out)
 	}
-	for _, want := range []string{
-		"app.kubernetes.io/component: cds",
-		"app.kubernetes.io/name: ratls-mesh",
-		"app.kubernetes.io/name: nri-image-policy",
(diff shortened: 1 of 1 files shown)


AmeanAsad (approved):
huge nice on this

PR #172, opened 2026-07-29
feat(secrets): serve /secrets from CDS

CluEleSsUK:
Wires secret release into CDS. `GET`/`POST /secrets/*` releases a secret to a
workload whose sandbox matches an allowlist entry holding a grant.

## What decides a release

1. The client leaf chains to the mesh CA — verified by crypto/tls, not the
   handler. The RA-TLS path is not accepted: it would admit a self-signed peer
   whose sandbox-ID extension is whatever it chose.
2. The sandbox ID comes from that verified leaf.
3. A fresh inventory-signed token carries this request's challenge and names the
   same sandbox.
4. The inventory bound to that sandbox reports what it has run.
5. That set, minus c8s's injected containers, matches exactly one entry.
6. The entry's grant covers the path.

Any failure refuses, as does an unreachable inventory, an unknown sandbox, an
empty container set, an ambiguous match, or an entry with no grant.

## Which inventory is asked

`/attest` records `sandbox ID → inventory host`, first-write-wins, and the
secrets path dials that host — the token must name the same one. A token is
signable by anything that can answer on the inventory port inside the configured
CIDRs, so taking the host from the request would let such a thing sign for
someone else's sandbox, be believed about what it runs, and have the answer
release its secrets.

The write lands after `verifySandboxWorkload` and before `SignCSR`, so a request
failing a later gate leaves no claim on a sandbox ID it never got a certificate
for. A conflict does **not** refuse the certificate: `get-cert` has no
token-less retry, so denying at issuance would let one pre-claim wedge a pod for
a whole certificate lifetime. The sandbox becomes unusable for secrets instead —
one workload rather than every pod on the node.

The ledger is built regardless of whether secrets are served, so a cluster that
starts serving them later does not begin with an empty one and refuse every pod
until its certificate renews.

## What turns it on

Writing a grant. An entry without one releases nothing, so the endpoint is inert
on a cluster whose entries hold none.

Serving it at all needs what sandbox identity already requires —
`--ratls-platform`, `--measurements`, `--sandbox-inventory-cidr`. Miss any and
CDS warns, naming the one it is missing, and does not serve `/secrets`. Nor does
it under handoff: a roll puts two CDS pods behind the Service, and the surge
replica serves an empty store, so a workload landing on it mints a value
diverging from the one its siblings hold. Not serving beats that divergence, and
beats failing startup, which would strand existing handoff installs.

## The injected drop set

c8s's own containers are removed before matching, so an entry never enumerates
them. A container is dropped when its digest is an allowlist **floor** entry
*and* its entrypoint is one c8s injects. Both halves matter: floor membership
alone would let a pod add busybox running a shell — also a floor entry — and
have it ignored.

The floor is the source because it already carries the injected image (it could
not run otherwise) and is additive, so an image bump leaves the previous digest
in place beside the new one and pods still running the old image keep matching
while they recycle.

This rests on no floor image other than c8s's having an executable at one of
those entrypoints. Floor contents are operator-controlled and auditable, but
that is a property of the deployment rather than something enforced in code.

## Routing, listener, policy

The challenge is `POST /secrets`, which the wildcard can never match, so it
shadows no store path — a secret at `/challenge` still resolves. Its challenge
pool is separate from issuance's, so a nonce minted for one is not redeemable at
the other. Paths must arrive canonical and are rejected rather than repaired, so
the bytes matched against a grant and the bytes used as a store key are the ones
the client sent.

The listener sets `ClientCAs` to the mesh CA with `VerifyClientCertIfGiven`, so
callers with no certificate reach every other route unchanged.

`Store.Version` is a single-row read the secrets path uses to decide whether its
cached allowlist is current; `LoadAll` is a full scan plus a JSON unmarshal per
entry and cannot sit on a request.

## Known gap

The secrets routes take the per-source-IP rate limiter, but pods reach CDS
through a NodePort and the host-network mesh proxy and collapse into one bucket,
so a hostile pod can exhaust its node's budget and wedge co-tenants. Keying on
`(source IP, sandbox ID)` needs the sandbox ID verified before the limiter runs.

## Not in this PR

The injected fetcher, its webhook wiring, and the chart values — a workload must
call the API itself for now.

See `docs/secrets.md`.


internal/cmds/cds/cmd.go
```diff
@@ -77,6 +77,12 @@ func NewCmd() *cobra.Command {
 	flags.DurationVar(&cfg.rateLimiterEvictInterval, "rate-limiter-evict-interval", time.Minute, "interval for per-IP rate limiter eviction sweep")
 	flags.DurationVar(&cfg.rateLimiterIdleTimeout, "rate-limiter-idle-timeout", 5*time.Minute, "idle duration before a per-IP rate limiter entry is evicted")
 
+	flags.BoolVar(&cfg.secrets, "secrets", false, "serve GET/POST /secrets/* — release secrets to a workload whose sandbox matches an allowlist entry holding a grant. Requires --measurements, --sandbox-inventory-cidr, --ratls-platform and --injected-component-digest, and is refused alongside --handoff-peer-url (docs/secrets.md)")
```
CluEleSsUK: handoffs in this model are definitely a problem hmm

PR #168, opened 2026-07-29
refactor(ratls): remove cert claims and replace with sandboxId digests

CluEleSsUK:
- removed cert claims of digests that only bind after renewal
- added API to nri-image-policy/policy-monitor to host `/digests/{sandboxId}` for the CDS to consume
- rejigged the verification logic to use the new sandboxId
- did a sweep over docs for consistency

PR #174, opened 2026-07-29
feat(cds): derive the sandbox-digests bound from the live node list

CluEleSsUK:
c8s install baked one /32 per node into cds.sandboxInventoryCIDRs at install time. The snapshot went stale on every scale-up: workloads on new nodes got certificates with no sandbox ID and no issuance-time image gate until someone re-ran install or helm upgrade (which also rolls CDS, the cluster trust root). The node lister became load-bearing when #168 made the bound gate every sandbox-token issuance.

With --sandbox-inventory-cidr unset, CDS now runs a node informer and derives the bound itself — one host route per InternalIP, same filters as the install-time inference (global unicast only; an address inside a pod range is excluded and logged). Add/update/delete events recompute the set, so scale-up is covered without a restart and removed nodes leave the bound (the snapshot never shrank either). An unsynced or empty set still refuses every sandbox token.

Explicit CIDRs keep the old static behavior exactly: the chart renders the node-reader ClusterRole and mounts the API token only when the value is empty. The trust delta is small by construction: the bound was already control-plane-supplied (as argv), Node objects are writable only by kubelets, and the set still only selects a dial target — what answers must pass mutually-attested RA-TLS on a privileged port.

c8s install keeps the kubectl read as a preflight (fails on pod-range overlap or an unreadable cluster) but no longer renders the value. InventoryHosts is now an interface with two sources: CIDRHosts (static) and NodeHosts (informer-swapped atomic snapshot).


b0xtch:
Attaching CI evidence that this is currently red on `main`.

At `05065c6a` the snp-metal e2e fails with:

```
attestation failed: server returned 403: {"error":"csr_denied",
"message":"sandbox token presented but no inventory CIDRs are configured to bound the callback"}
```

https://github.com/confidential-dot-ai/c8s/actions/runs/30559385697

The CVM boots and passes `assert genuine confidential boot`. What fails is issuance: `demo-nginx` sits `0/1` for ten minutes because `c8s-cert-wait` never gets a certificate, and get-cert then enters the 6h renewal loop rather than failing fast.

The bracket lines up with #168:

- last green scheduled e2e: 2026-07-29 09:17Z
- eef5c5d4 (#168) merged 09:36Z
- every run since: the same 403

It surfaces here because this lane installs the chart from source into an rke2 `HelmChart` CR with inline `valuesContent` instead of going through `c8s install`, so `cds.sandboxInventoryCIDRs` stays at the `[]` chart default. That is the state this PR makes work by deriving the bound from the live node list, so landing it should clear the lane without the workflow having to carry the value.

One note on blast radius, from reading the chart rather than from a captured log: `tls-lb` takes its certificate through get-cert init containers (`tls-lb-deployment.yaml:58`), so on a cluster in this state the denial should stop that pod from starting at all, not only workload pods.

Correcting my own first pass here: we do have a `tls-lb` at `Init:Error` on another lane running this commit, but that lane installs through `c8s install`, which populates the value, and it does not capture container logs. So I am not claiming the same cause there.

No change requested on the PR itself.

PR #175, opened 2026-07-29
feat(secrets): injection annotations

CluEleSsUK:
Two annotations - confidential.ai/c8s-secrets (NAME=/path list, reusing the existing listAnnotation helper) and confidential.ai/c8s-secret-dir. Both added to hasInjectionDetailAnnotations, so setting either without confidential.ai/cw errors rather than being silently ignored.

The fetcher injects as a native sidecar ordered c8s-cert → c8s-cert-wait → c8s-secret, so the leaf it authenticates with is already on disk. There's a test pinning that order and the restartPolicy: Always.

Guards, each mirroring shipped cert-path precedent:

- rejectReservedSecretsVolume — ensureVolume keeps a pre-declared same-named volume, so without this a pod spec pointing c8s-secrets at a hostPath would have CDS-released secrets written outside the TEE. Table test covers hostPath, disk-backed emptyDir, and PVC.
- c8s-secret added to isReservedCertName — covers both regular and (existing) ephemeral-container checks.
- RW for the fetcher, RO for everyone else. This works because mountAll skips a container that already has the mount, so ordering the pod-wide RO mount before building the fetcher leaves its own RW mount intact. Worth knowing that's load-bearing, not incidental.
- sizeLimit on the tmpfs — it's charged to the pod's memory (the guest's, under kata).

One new Config field: CDSMeasurements, rendered into the fetcher's --measurements. Empty pins nothing, which leaves an impostor CDS able to answer with a value of its choosing — get-cert has the same posture today, but it matters more here, so it's documented on the field.

Handler — Handle now branches on req.SubResource == "ephemeralcontainers" before anything else. That request shape is different: the pod already exists, nothing is injected, and the only decision is whether the new containers may reach what the injected ones put there. rejectEphemeralReservedMounts denies any ephemeral container mounting the secrets volume or the pod's cert volume, and any claiming a reserved container name.

The cert volume name is derived from the pod's own c8s-cert-volume annotation, not a constant — a pod that renamed its cert volume must have that name protected, and the default name is then no longer sensitive. There's a test for both directions.

Chart — added a second rule to the mutating webhook for pods/ephemeralcontainers on UPDATE. Confirmed with helm template that it renders; the whole rule block is above.

**Caveat:**
I did not add pods/ephemeralcontainers to the deny-host-namespaces VAP, which the earlier review had suggested. Its CEL targets host namespaces and hostPath volumes, and an ephemeral container can introduce neither — it can only mount volumes the pod already declares. So the mutating webhook is the correct place, and widening the VAP would have meant re-reasoning about CEL that doesn't apply.

Also worth noting the guard protects the cert volume too, not just secrets. That gap predates this work — kubectl debug mounting c8s-certs could read the pod's mesh private key  — and closing it here was free.

### DIFF
--- a/internal/webhook/pod_mutator.go
+++ b/internal/webhook/pod_mutator.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/confidential-dot-ai/c8s/pkg/workloadclaims"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/util/validation"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/log"
@@ -73,6 +74,13 @@ const (
 	AnnotationGetCertRunAsGroup      = "confidential.ai/c8s-get-cert-run-as-group"
 	AnnotationGetCertRunAsNonRoot    = "confidential.ai/c8s-get-cert-run-as-non-root"
 	AnnotationGetCertVerbose         = "confidential.ai/c8s-get-cert-verbose"
+
+	// AnnotationSecrets requests secrets for the pod, as a comma-separated
+	// list of NAME=/store/path. NAME is the file each value is written to
+	// under AnnotationSecretDir. Setting it injects the fetcher sidecar.
+	AnnotationSecrets = "confidential.ai/c8s-secrets"
+	// AnnotationSecretDir overrides where the files land.
+	AnnotationSecretDir = "confidential.ai/c8s-secret-dir"
 )
 
 var errInvalidInjectionAnnotation = errors.New("invalid c8s injection annotation")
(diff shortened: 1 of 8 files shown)


internal/webhook/pod_mutator.go
```diff
@@ -759,8 +823,20 @@ func mutatePod(pod *corev1.Pod, inj *injection, cfg Config) {
...
+		// so mountAll (which skips a container that already has the mount)
+		// leaves the fetcher's own read-write mount alone.
+		mountAll(pod, corev1.VolumeMount{
+			Name:      secretsVolumeName,
+			MountPath: effective.Secrets.Dir,
+			ReadOnly:  true,
+		})
+		injected = append(injected, secretContainer(&effective, cfg))
+	}
+	pod.Spec.InitContainers = injectInitContainers(pod.Spec.InitContainers, injected...)
```
CluEleSsUK, line 839: initially I wondered if this would block shit, but with `restartPolicy: Always` it's non-blocking and the main containers start np

PR #177, opened 2026-07-29
feat(cds): attested cluster governance — CDS-cert config-claims, live-allowlist reissue, RTMR pinning, argv-aware workload stamp

AnomalRoil:
## Why

An external client verifying this cluster today bottoms out at a PEM someone handed it: the mesh CA is pinned out of band, and nothing attested says what admission policy is in force. #168 correctly removed per-leaf cert claims (they only rebind at renewal) in favor of live sandbox-ID digests — but that also removed CDS's own governance attestation, leaving no attested channel at all for the mesh CA, the operator key set, or the allowlist toward clients that cannot reach the inventory (browsers, external proxies).

This PR restores that channel in the place where #168's staleness objection does not apply: **CDS's own self-signed serving certificate**, which is reissued the moment the state it attests changes. Issued workload leaves are untouched — workload identity stays on the #168 sandbox-digests path, and this PR strengthens it.

## What

1. **Config-claims on the CDS serving certificate only** (`pkg/ratls/claims.go`, OID .1.3): operator-keys digest, seed digest, **mesh-CA digest** (SHA-256 of the issuing CA DER), and the **live allowlist digest** (canonical bytes of exactly what `GET /allowlist` serves), DER-bound into the quote's REPORTDATA. The wire format is the established v3 encoding — two external verifiers (TEErminator, c8s-verify-js) already parse it byte-for-byte; the legacy workloadDigest field rides along always-unset for compat.

2. **Reissue-on-change** (`internal/cmds/cds/allowlistclaims.go`): the serving cert is swapped whenever the live allowlist digest moves, so the certificate fingerprint is the cache-invalidation signal — clients cache a verified attestation indefinitely and re-attest exactly when policy changes. The watcher gates hashing on the store version and retries failed swaps without advancing state.

3. **Discovery republish** (`discovery.cds_identity`): get-cert records the CDS certificate it verified during issuance (`OnVerifiedPeer` fires only after attestation passes) and republishes it in the discovery document, so a browser can attest CDS without reaching it. Safe over an untrusted path: the certificate is self-authenticating (evidence binds its own key and claims).

4. **Verify CLI**: `--mesh-ca` / `--mesh-ca-digest`, `--allowlist` / `--allowlist-digest`, `--allowlist-seed[-digest]`, plus runtime-register pinning — `--expected-rtmr1/2/3`, `--operator-key` (RTMR[3] = operator-key binding), and `--image-manifest` (pins MRTD+RTMR1+RTMR2 together from the confos manifest, since MRTD alone covers only firmware — two different guest images built on one OVMF share an MRTD). `pkg/rtmr3` generalizes to `pkg/runtimemeasure` as the single home of the extend conventions.

5. **Argv-aware matched-workload stamp** (OID .1.5): at issuance CDS now fetches per-container `(digest, argv)` from the inventory (`FetchSandbox`, #170) and stamps the allowlist entry (or entries) the sandbox matches, reusing the #170/#171 argv-policy matcher. This restores — strictly stronger — the entry-combination gate #168 dropped: non-floor containers matching no entry refuse issuance. Two entries sharing an image digest and differing only in argv are now distinguished by the argv the inventory observed; the stamp is marked ambiguous only when argv genuinely cannot separate them. Inventories predating per-container detail degrade to the flat digest gate with no stamp.

## Compatibility

- Wire-identical claims + REPORTDATA transcript to the external parsers (verified: zero code drift against their reference).
- Older inventories: flat-digest gate unchanged, no stamp, no refusal.
- Pre-#168 leaves read via `--from-file` still surface their workload digest.
- `--mesh-ca` is dual-semantic by kind: claims-digest pin for `kind=cds` (self-signed target), leaf-chain check otherwise (retains the #168 sandbox-ID semantics).

## Testing

`go test ./...` and `-race` green; `golangci-lint` 0 issues. Ported and adapted the full test surface from the two predecessor branches plus new coverage: reissue-on-change and swap-failure retry; claims v1/v2/v3 parse/pin matrices; argv disambiguation (same digest, different exact argv → single entry), both-`any` ambiguity, off-policy argv refusal, no-detail degradation. The predecessor of this chain was exercised end-to-end against a live bare-metal TDX cluster (mesh-CA + allowlist pins in both directions, RTMR pins matched to the confos manifest and hardware quote).

Supersedes `feat/cds-rollup` and `feat/cds-stamp-matched-entry` (pre-#168 architecture). Client-side counterparts: c8s-verify-js #17 + its stacked caching branch; TEErminator `feat/cds-rollup`/`feat/cds-identity-cache`. Docs (`docs/ratls.md` config-claims section) still describe the pre-#168 world and need a separate pass.

**History note:** `feat/cds-rollup` (@AmeanAsad) is grafted into this branch's ancestry via a `-s ours` merge — content and authorship unchanged, re-signed to pass the signed-commits gate — so the roll-up's attribution survives the squash.


AnomalRoil:
Review round complete (full report kept internally; approve-with-minor-comments). All actionable findings are now addressed in `702b582`:

- **Argv-skew degradation**: a non-floor container the inventory reports with detail but empty argv (admitted before argv recording — mixed-version upgrade window) now degrades to issue-without-stamp instead of failing exact-argv policies and refusing renewal of an admitted pod. Floor containers' missing argv remains irrelevant. Test: `TestAttest_SandboxWorkload_EmptyArgvDetailIssuesWithoutStamp`.
- **`--mesh-ca` claims-pin mode refuses multi-block PEMs** rather than silently digesting block one (a rotation-era bundle would have read as an attack). Test added.
- **Stale trust-semantics header** in `pkg/ratls/matchedworkload.go` rewritten for the argv-aware match (it still described the pre-#168 digest-only design).
- **Doc drift**: `docs/ratls.md` OID table now lists .1.3 (CDS serving cert only) and .1.5 instead of claiming .1.3 retired; `docs/kata-guest-base.md` references `pkg/runtimemeasure`.
- **Comment accuracy**: operator-keys cross-check wording no longer claims 'mandatory' where `--from-file` targets structurally can't run it.

Deliberately NOT changed, for the record: (a) mixing containers from different entries refuses issuance again — that is the point of restoring the combination gate, and the flat-digest/floor-only/no-detail degradation paths are all tested; (b) `c8s verify --kind cds --operator-keys` against a claims-free pre-this-PR CDS fails closed — upgrade the server before the CLI, or drop the pin during transition; (c) `pkg/rtmr3`'s deletion is a Go API break, checked against known external consumers (none import it — TEErminator and c8s-verify-js reimplement the wire formats deliberately); (d) `cds_identity` freshness stays bounded by the serving-cert TTL — consumers must treat the fingerprint as a cache key and cross-check the live allowlist bytes, which both client implementations do.

Full suite + `-race` + golangci-lint clean at `702b582`; trial-merge against current main was clean at review time.

internal/cmds/cds/attest.go
```diff
@@ -308,61 +313,139 @@ func (h AttestHandler) verifySandboxToken(ctx context.Context, raw json.RawMessa
...
+//
+// Floor digests are excluded: they are admitted alone and carry no combination
+// policy, and c8s's injected sidecars (get-cert) are floor entries, so their
+// digest drops out here. A floor-only pod matches no entry and is stamped with
+// nothing. Ambiguity — more than one surviving entry — is a real outcome (two
+// entries can admit the same commands, e.g. both with an "any" argv policy)
+// and the stamp names the whole set rather than asserting an identity the
+// evidence does not establish. Zero surviving entries with non-floor
+// containers present refuses issuance.
+func (h AttestHandler) matchWorkloadEntries(containers []workloadclaims.SandboxContainer) (*ratls.MatchedWorkload, error) {
```
CluEleSsUK, line 410: I was gonna say 'do we not check init vs main' here, but I guess there is no way to know from sandbox :<

CluEleSsUK: seems like a fairly niche threat area though

internal/cmds/getcert/run.go
```diff
@@ -205,15 +205,19 @@ func run(cfg config) error {
 	}
 	slog.Debug("output paths validated")
 
-	client, err := newCDSClient(cfg)
+	// Record the CDS certificate this process verifies during issuance, so the
```
CluEleSsUK, line 208: seems like the comment should be where it actually happens rather than obj creation

internal/cmds/getcert/run_more_test.go
CluEleSsUK, line 1: nice test name haha

internal/cmds/getkubeconfig/cmd.go
```diff
@@ -53,6 +53,11 @@ func NewCmd() *cobra.Command {
...
 	f.StringVar(&cfg.APIServerURL, "apiserver-url", "", "apiserver URL for the kubeconfig (overrides --node)")
 	f.StringVar(&cfg.OperatorKeyPath, "operator-key", "", "operator ECDSA private key PEM (its public half is bound into RTMR[3]) (required)")
+	f.StringVar(&cfg.Image.ManifestPath, "image-manifest", "", "confos build manifest.json for the guest image the node must be running; pins tdx.mrtd, tdx.rtmr1 and tdx.rtmr2 together (oras pull ghcr.io/…/c8s-base:<tag>). REQUIRED unless the three are given individually — RTMR[3] is derived from your PUBLIC key and the host stages it at launch, so without an image pin any TDX guest can satisfy the gate")
+	f.StringVar(&cfg.Image.MeasurementHex, "measurement", "", "expected MRTD as 96 hex chars, if you have the digests rather than the manifest. TDX MRTD covers only the firmware, so it must be given with --expected-rtmr1 and --expected-rtmr2")
+	f.StringVar(&cfg.Image.RTMR1Hex, "expected-rtmr1", "", "expected RTMR[1] as 96 hex chars — the guest kernel (UKI image identity + GPT + boot)")
+	f.StringVar(&cfg.Image.RTMR2Hex, "expected-rtmr2", "", "expected RTMR[2] as 96 hex chars — the guest rootfs (UKI section measurement chain)")
```
CluEleSsUK, line 59: should we enforce in this file that if `expected-rtmr1` is set then so is `rtmr2` or are there reasonable scenarios for only setting 1?

AnomalRoil:
This PR is being retired in favour of a better, cleaner verification refactoring done in #197 , #199, #230 and #245

PR #181, opened 2026-07-29
feat(tls-lb): wide-open CORS on protocol endpoints by default

AnomalRoil:
## Why

Browser verifiers are dead on arrival against a default install. The protocol endpoints — `/.well-known/c8s/*`, the discovery document and certificate endpoints, and the built-in `/allowlist` route — send no CORS headers unless the operator enables the global `tlsLb.cors` block, which also requires choosing an origin allowlist and applies to every workload location. So the endpoints whose whole purpose is *any browser anywhere can verify this cluster* are exactly the ones a cross-origin page cannot reach (we just hit this deploying the static c8s-verify demo page, and had to add CORS at the WebPKI front door instead).

An origin allowlist on those endpoints protects nothing:

- every response is self-authenticating (hardware evidence, CDS-signed certificates, sealed tunnel records) or public by design;
- no request relies on ambient browser credentials — allowlist mutations are operator-signed over method, path, and body;
- the client-side c8s-verify flow fails closed on substitution.

It only breaks third-party verifiers — and third parties verifying your cluster from origins you never heard of is the point of publishing attestation.

## What

New `tlsLb.cors.protocolEndpoints` (default **true**): the protocol-owned locations serve `Access-Control-Allow-Origin: *` (no credentials, preflight short-circuited at nginx) via a self-contained directive set — no http-level maps and no upstream pass-through, since these endpoints are c8s-owned end to end.

Scope and precedence:

- **Workload locations** (`routes` and the catch-all upstream) are never affected by this flag; they keep the opt-in global block.
- **An enabled global `cors` block behaves exactly as today** and covers the protocol endpoints with the operator's policy; the built-in wide-open block steps aside (no double headers, no behavior change for existing CORS users).
- `protocolEndpoints: false` restores the previous no-CORS default.
- Preflight `OPTIONS` on `/allowlist` returns 204 before the rate limiter, which already exempted it.

## Testing

- New chart tests: default render carries `ACAO *` on all six protocol locations and nothing on the catch-all; opt-out renders no CORS at all; an enabled global block uses the `$cors_out_*` machinery on protocol locations with no literal `*` anywhere; `--set-string` booleans are rejected.
- `go test -race ./...`, `golangci-lint run` (0 issues), `gofmt`, and `.github/scripts/chart-verify.sh` all clean.

### DIFF
--- a/internal/helmchart/c8s/templates/tls-lb-configmap.yaml
+++ b/internal/helmchart/c8s/templates/tls-lb-configmap.yaml
@@ -211,6 +211,8 @@ data:
             location = {{ .Values.tlsLb.discovery.path }} {
                 {{- if default false .Values.tlsLb.cors.enabled }}
                 {{- include "tls-lb.corsLocationDirectives" .Values.tlsLb.cors | nindent 16 }}
+                {{- else if eq (include "tls-lb.protocolCorsEnabled" .) "true" }}
+                {{- include "tls-lb.protocolCorsLocationDirectives" . | nindent 16 }}
                 {{- end }}
                 access_log off;
                 default_type application/json;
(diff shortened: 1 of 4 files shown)


joaosa (approved):
lgtm

PR #194, opened 2026-07-30
Fix/tdx sandbox identity and volumed

CluEleSsUK:


internal/cmds/nri-image-policy/config.go
```diff
@@ -175,6 +205,18 @@ func (c *config) AllowlistEnabled() bool {
...
 func (c *config) Validate() error {
+	// Reject at load rather than at the first CDS handshake: a wrong platform
+	// produces a peer-attestation failure on the CDS side that names the
+	// evidence platform, not this setting, so the cause is several hops from
+	// the symptom.
+	// Compare the NORMALIZED form: NormalizePlatform folds "snp" onto
+	// "sev-snp", which is what ratls.parseTEEType accepts downstream, so
+	// checking against types.PlatformSnp here would reject the default.
+	switch c.NormalizedPlatform() {
+	case "sev-snp", string(types.PlatformTdx):
```
CluEleSsUK: we should probably change this in `ratls.parseTEEType`
also the switch here is harder to read than an `if`

CluEleSsUK:
squashing with signing

PR #227, opened 2026-08-04
ci(tdx-metal): vendor the tool-fetch retry from confidential-ci#59

b0xtch:
confidential-ci#59 added a retry around the tools step's four internet fetches. This is the vendored copy, and its own header requires the two to stay byte-identical:

```
# workflow_call so the vendored copy can be invoked as a job by c8s's
# e2e-c8s.yml. workflow_dispatch so either copy stays runnable by hand.
# Both copies carry both, so the two files remain byte-identical.
```

Taken verbatim from confidential-ci main after that merge. Blob SHA `02f99dc8` on both sides, so they are identical again.

The change: that step fetches kubectl, go, crane and helm over the public internet under `set -e`, so one hiccup fails the lane before it reaches a CVM. Twice in six days on the other copy:

```
30529875545  2026-07-30  proxy.golang.org: HTTP/2 stream error INTERNAL_ERROR
30895640537  2026-08-04  lookup proxy.golang.org on 172.21.0.10:53: server misbehaving
```

Each fetch now gets 3 attempts with 5s/10s backoff. Also fixes dead code: `GOVER="$(curl ... | sed ...)"` under `set -euo pipefail` exits on a failed fetch, so the pinned `1.26.3` fallback was unreachable (verified: exits 6 as written, reaches the fallback with `|| true`).

Nothing enforces the byte-identity invariant in CI, which is why the drift went unnoticed between #59 merging and this PR. Worth a check if the invariant is meant to hold.

### DIFF
--- a/.github/workflows/tdx-metal-e2e.yml
+++ b/.github/workflows/tdx-metal-e2e.yml
@@ -374,10 +374,20 @@ jobs:
         run: |
           set -euo pipefail
           rm -rf /tmp/c8s-src
-          git clone -q https://github.com/confidential-dot-ai/c8s.git /tmp/c8s-src
+          # Same exposure as the tools step: this clone and the module fetches
+          # behind `go install` cross the public internet under `set -e`.
+          for i in 1 2 3; do
+            git clone -q https://github.com/confidential-dot-ai/c8s.git /tmp/c8s-src && break
+            [ "$i" = 3 ] && { echo "::error::git clone failed after 3 attempts"; exit 1; }
+            echo "clone attempt $i failed, retrying in $((i * 5))s" >&2; rm -rf /tmp/c8s-src; sleep $((i * 5))
+          done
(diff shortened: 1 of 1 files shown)


.github/workflows/tdx-metal-e2e.yml
```diff
@@ -106,14 +106,29 @@ jobs:
       - name: tools (kubectl, go, crane, helm)
         run: |
           set -euo pipefail
+          # Each fetch below crosses the public internet from an ephemeral ARC
+          # pod; one hiccup killed the lane twice in six days (30529875545,
+          # 30895640537), both before it reached a CVM.
+          retry() {
```
joaosa: It would be a nice to have an exponential backoff primitive which could be reused for cases like this

b0xtch: done

PR #230, opened 2026-08-04
feat(verify): platform-complete measurement policy and certificate-body authentication

AnomalRoil:
Verification refactoring, stacked on #199. Three related hardenings of the operator-side verifiers.

## One shared runtime-measurement package
`pkg/runtimemeasure` replaces `pkg/rtmr3` (same convention — the golden vectors are ported verbatim and a failure is a breaking change to the attestation contract). It adds what was missing or duplicated elsewhere:
- `ForOperatorKey` — the operator-key RTMR[3] seed, previously duplicated byte-for-byte in `get-kubeconfig` and `cred-release`;
- `FromDigestsSeeded` — the seeded, ordered workload-event chain;
- `CanonicalDigest` — strict `sha256:<64-lower-hex>` canonicalization (digest-pinned refs accepted, tags rejected);
- `LoadImageManifest` — a strict parser that loads the TDX image tuple **MRTD + RTMR[1] + RTMR[2]** atomically from one explicitly selected build-artifact manifest (all three fields required, 96 lowercase hex each; a generic artifact-hash `manifest.json` is rejected with an explanation). `rtmr3-measurer` migrates to the new package.

## `c8s verify`: platform-complete TDX policy
- `--image-manifest`: MRTD joins the `--measurements` allowlist; RTMR[1]/RTMR[2] are pinned exactly — the guest kernel and rootfs are verified rather than only the TDVF firmware.
- `--expected-rtmr3`: pins the runtime operator-key/workload-event chain (a deployment property, not a cluster identity).
- **Any RTMR pin against non-TDX evidence is a hard policy error**, never an ignored option; absent/malformed claims fail closed.
- A TDX verdict pinned on MRTD alone now warns prominently that the policy leaves the kernel/rootfs unmeasured.

## Certificate-body authentication on every cert-sourced path
The RA-TLS attestation extension binds only the public key, so previously any other field of a certificate — subject, serial, validity, extra extensions — could be rewritten under a genuine attestation and `ratls-cert`/`discovery`/`--from-file` verification would still pass, with no expiry check at all. Now (`pkg/certutil.AuthenticateLeafBody`, shared with get-kubeconfig):
- self-issued leaves must verify their own signature with their attested key before any body field is used;
- `NotBefore` is enforced with a single documented 5-minute skew, `NotAfter` with none, on every path (the new attest-pq client already did this — the same constant now backs all of them);
- the verdict reports what authenticates the body (own attested key / transcript-committed CA / `--mesh-ca` chain / CA-vouched-unauthenticated), and nonce-free paths state that freshness comes only from the validity window;
- discovery mode now retains the leaf, so `--mesh-ca`/`--sandbox-id`/`--workload` pins work there too.

## `get-kubeconfig` trust gate
The gate used to check only platform=TDX and `rtmr_3 ==` the bare operator-key seed. That is forgeable — the untrusted host stages the operator key on the `opkeydata` disk, so it can boot **any** guest image and reproduce that register — and it also breaks on any node whose measurer has extended a workload event. Now:
- `--image-manifest` is **required**: the image tuple is the identity anchor;
- expected RTMR[3] = `FromDigestsSeeded(ForOperatorKey(pubPEM), ordered --workload-image digests)` (bare seed only when no workload images are expected);
- the full policy is enforced **identically** on the initial attestation gate and the RA-TLS credential-release connection through one shared funnel, which also pins the single-wrap evidence-envelope shape (a double-wrapped envelope is refused loudly) and authenticates the presented certificate body;
- `cred-release` derives its startup expectation from the shared helper; its bare-seed equality is documented as valid because the node image runs no workload measurer.

## Tests
Golden vectors (convention unchanged + `ForOperatorKey` + seeded chain); manifest reject tables (missing/malformed/mixed fields); RTMR pins on SNP rejected; MRTD-only TDX warning; altered self-signed body, invalid self-signature, future `NotBefore` beyond/within skew, expired `NotAfter`, replay-bounded-by-validity; get-kubeconfig gate matrix incl. wrong tuple, wrong chain, workload-extended register, double-wrapped envelope. `go test ./...`, `golangci-lint run` (0 issues), gofmt all green.

Docs updated in the same PR: THREAT_MODEL RTMR rows, ratls.md known-gap, operator.md TDX pin statement + a new `get-kubeconfig` trust-gate section, kata-guest-base.md package references.


AnomalRoil:
## Hardware validation on live Intel TDX (b300-node-1)

Ran the full stack (this branch @ 45a03e2, i.e. #197+#199+#230, plus c8s-verify-js#20 and TEErminator#15) on a genuine TDX trust domain: KubeVirt CVM (`launchSecurity: {tdx: {}}`, qemu `-object tdx-guest`), root disk CDI-imported from the published `c8s-base:rke2-cdi-789f99a`, branch component images built locally and served from an in-CVM registry (allowlisted by digest via the operator key — itself a live exercise of the allowlist path).

| Check | Result |
|---|---|
| Two-entry allowlist (base/extended pair) discriminates; unnamed→named at first fast renewal (observed t+36s, 30s+jitter) | PASS |
| Verified front door not-ready (503, endpoints withheld) until the mesh leaf is stamped, then ready | PASS (after the chart fix below) |
| `c8s verify --workload/--allowlist` verdict matrix (verified/absent/name-mismatch/digest-mismatch/usage errors, exit codes) | PASS |
| `--mode attest-pq` deployment-class and specific-cluster verdicts; `--image-manifest` tuple matched live; one-nibble RTMR[2] edit rejected; MRTD-only warning verbatim | PASS |
| Endpoint contract: old `/attestation` → 400 `invalid_request`; `?pq=`/`?binding=` → 400; attest-pq+attest-lb 200 on cds mode; live WebPKI flip → `attest-lb` 400 `unsupported_front_door`, attest-pq still served | PASS |
| TEErminator attest-lb: Verified + workload + trust mode in `status`; wrong pin → 502, nothing forwarded; leaf rotation → re-attest on next handshake | PASS |
| c8s-verify-js live: both trust classes verify, stamp surfaced, sealed tunnel round-trips; typed failures (`workload_denied`, `measurement_denied`, `allowlist_denied`) | PASS |
| CDS restart w/o handoff: new CA; old-CA pin rejected (`unknown authority`), deployment-class re-accepts after fresh attestation | PASS |
| get-kubeconfig gate: missing manifest → usage error; wrong RTMR[1] → refused naming the register; wrong operator key → RTMR[3] chain mismatch; correct tuple → `kubectl auth whoami` = `operator` | PASS |
| Secrets release unchanged throughout | PASS |

Observed node measurements matched the oras artifact's manifest exactly (MRTD `9309eaae…8ba1`, RTMR[1] `d3e7d563…2222`, RTMR[2] `43064155…d676`).

**Fixes that came out of this run:**
- `d90184c` (on #199): `tlsLb.attest.expectedWorkload` was unsatisfiable — the tls-lb get-cert had no `--workload-claims` and no inventory socket, so the gate held the front door at 503 forever. The chart now wires the claims flow (socket mount + supplemental group 65532 on node-CVM, guest loopback under kata) whenever the gate is enabled.
- `9eba0c4` (here): the tdx-metal-e2e lane now supplies the image tuple (`rtmr1`/`rtmr2` in the refs CM → `--image-manifest`), which this branch makes mandatory for get-kubeconfig.
- #233: workload matching drops every container whose argv[0] is `/c8s` (incl. cds-attest/allowlist-proxy) — design decision needed, filed separately.

Not proven in this run: rejection of a *relayed* attest-lb response behind a different live leaf (only the rotation path was forced), and the manifest values were read from a verified quote rather than an independently provenanced build artifact — the provenance pipeline gap is known.

PR #256, opened 2026-08-06
feat(cds): external KMS backend for the secret store, Azure Key Vault…

CluEleSsUK:
… first

A store path can now be backed by an external KMS: CDS fetches the value at release time and forwards it to the attested workload, instead of holding it in process memory. This is the direct-brokering shape of whitepaper §5.6.4, with one deliberate substitution — Key Vault cannot verify CDS's attestation (SKR is locked to MAA, which attests Azure-hosted TEEs only), so CDS authenticates with an operator-provisioned Entra credential.

The release gate is untouched: mesh leaf, challenge, sandbox token, exact workload match, grant. Only the store lookup changes. The Store interface was written for this ("a KMS wraps a key, a vault mints its own"); RoutingStore dispatches mapped paths to the KMS and everything else to the memory store. The webhook, get-secret sidecar, and grant model carry no new code. Brokering rather than `secrets put` because the vault stays the custody and audit point, vault-side rotation takes effect on pod restart with no operator action, and plaintext keys never transit the operator workstation.

The operator connects the KMS with `c8s secrets external apply` over the existing attested channel (RA-TLS pinned to the CDS measurement, operator JWT bound to method/path/body). The credential lives in CDS memory only — never a Kubernetes resource. The mapping set (paths, vaults, secret names; no credential) persists beside the allowlist DB, so a restarted CDS fails mapped paths closed instead of minting a divergent value; re-applying re-arms them. That file is host-visible, so the post-restart runbook — `c8s cds verify --operator-keys`, then re-apply — is the detector for a swapped key set or a rolled-back mappings file, not a prevention.

One backend at a time; the document's `backend` field names it so future backends share PUT/GET /secrets/external. `/external` is a reserved store path. Mapped paths are read-only from c8s: workload POST 409s without minting, operator PUT 409s naming the external backing. Fetches are versionless (latest at request time), values are delivered verbatim, fetch failures are the same opaque 500 as any store error, and the get-secret sidecar's retry loop is the only retry.

Stated residuals: the Entra credential is bearer material — a CDS compromise exposes it and in-flight values (the accepted direct-brokering trade-off; RBAC scope is the blast-radius bound, and the cached Entra tokens are bearer too). A control plane booting a same-measurement clone could harvest it from a steered apply; that inherits the known clone gap in docs/operator.md. SKR remains the upgrade that lets the KMS itself check CDS's attestation, viable once CDS-on-Azure is a supported shape; gaps.md's close criterion stays open until then.

Verified by unit and handler tests plus an in-process end-to-end test against a stubbed Entra+vault TLS server: real operator-JWT apply, attested workload fetch through the full gate, no-mint and operator-write refusals, grant-first denial, restart fail-closed, and re-apply recovery. `go test -race` clean on the touched packages. Not run against a live Azure tenant or on hardware.

### DIFF
--- a/docs/operator.md
+++ b/docs/operator.md
@@ -246,7 +246,10 @@ CDS verifies EAR JWTs against its own in-process signer; there is no JWKS
 fetch to a separate component. The chart does not render a CA private key into
 a Kubernetes Secret. CDS generates its mesh CA key inside the process, keeps it
 in memory, and persists only the public CA bundle in the configured
-public-bundle PVC.
+public-bundle PVC. The same in-memory rule covers any external KMS
+credential an operator applies with `c8s secrets external` (docs/secrets.md): it
+arrives over the attested channel, lives in CDS memory, and is never rendered
+into a Kubernetes resource.
 
 Minimal allowlist-write values (pin operator public keys):
 
--- a/docs/secrets.md
+++ b/docs/secrets.md
@@ -190,7 +190,78 @@ keeps it until it is rolled. Replacing a path a workload created is worth
 pausing over for that reason: the pods that generated the value go on using it.
 
 Revoking a value means restarting CDS, which empties the whole store — see
-"Restarts".
+"Restarts". An externally-backed path is revoked at the KMS (delete or disable the
+secret, or cut the grant's access) and takes effect on the next fetch.
+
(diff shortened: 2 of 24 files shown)


CluEleSsUK:
old school, nuking

PR #262, opened 2026-08-07
test(ratls): fix TestProxyRunEndToEnd flake under coverage

joaosa:
## Problem

TestProxyRunEndToEnd fails the Coverage gate (main's latest CI run included) with `read: connection reset by peer`. The test's sequential round trips assume both semaphore slots are free once the client sees EOF, but the slots release in the handler goroutines' deferred cleanup, which runs after that. The next dial races the release; at capacity 2 the accept loop rejects the new connection and closes it with the payload unread, which surfaces as an RST on the client read. Coverage instrumentation widens the window enough to hit it every run.

## Fix

Wait for the semaphore to drain (`len(connSem) == 0`) before each round trip, using the test file's existing assertEventually helper. Test-only change; the proxy's release-after-teardown behavior is intended.

### DIFF
--- a/internal/cmds/ratlsmesh/proxy_serve_test.go
+++ b/internal/cmds/ratlsmesh/proxy_serve_test.go
@@ -355,6 +355,12 @@ func TestProxyRunEndToEnd(t *testing.T) {
 	})
 
 	for i := range 3 {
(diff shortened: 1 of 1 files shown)


joaosa:
this has been solved

PR #265, opened 2026-08-07
e2e(snp-metal): configure sandbox inventory CIDRs so CDS stops denying CSRs

joaosa:
The snp-metal cvm lane has failed on every run since #168 with `csr_denied: sandbox token presented but no inventory CIDRs are configured to bound the callback` — the demo workload's injected c8s-cert sidecar never gets its certificate, so it never reports PAYLOAD_OK (verified identical signature on yesterday's pre-v0.3.0 run, so unrelated to the confos bump).

Root cause: CDS fail-closes sandbox-token CSRs without configured inventory CIDRs (#168), and this lane writes raw chart values into a HelmChart CR — bypassing `c8s install`, which is what normally resolves and injects `cds.sandboxInventoryCIDRs` as per-node host routes. Same class of gap as the `deriveComponents` fix already documented in this file.

Fix, using the lane's existing placeholder+sed pattern (`RUNTIME_MEASUREMENT` precedent):
- values gain `sandboxInventoryCIDRs: ["NODE_CIDR"]`
- the in-guest payload substitutes the single node's `InternalIP/32` (exactly what `c8s install`'s `inferInventoryCIDRs` emits), with a `FAIL_NODEIP` marker if the lookup fails
- the runner-side render gate substitutes a dummy CIDR so `helm template` still validates the values

Tested via workflow_dispatch of `confidential-e2e` on this branch (run linked in comments once green).

### DIFF
--- a/.github/workflows/snp-metal-e2e.yml
+++ b/.github/workflows/snp-metal-e2e.yml
@@ -148,6 +148,8 @@ jobs:
                                   # resurfaces -> CDS Pending (run 29536583462).
             ratlsPlatform: sev-snp
             measurements: ["RUNTIME_MEASUREMENT"]
+            # NODE_CIDR placeholder: payload substitutes the node's InternalIP/32 (matching `c8s install`'s per-node host routes) — CDS fail-closes sandbox-token CSRs without it (#168; this lane writes raw values, so `c8s install` never fills it).
+            sandboxInventoryCIDRs: ["NODE_CIDR"]
           attestationApi:
             image:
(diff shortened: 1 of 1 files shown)


joaosa:
Validation dispatch: https://github.com/confidential-dot-ai/c8s/actions/runs/31171998401

joaosa:
Validation run is green across all three lanes: snp-metal/cvm reports PAYLOAD_OK for the first time since #168: https://github.com/confidential-dot-ai/c8s/actions/runs/31171998401

PR #290, opened 2026-08-11
feat(join): add heterogeneous node policy registry

aamirrasheed:
## Scope

Stacked on #267. This draft adds the first secure building block for heterogeneous c8s nodes: an immutable, platform-keyed admission policy for attested join/release.

It supports native TDX and native SNP evidence in the join protocol. It preserves the existing no-policy TDX flow.

## Security behavior

- The policy file must be on a read-only Linux mount.
- Evidence first selects a fixed policy profile; it cannot select an arbitrary verifier configuration.
- The local node and each peer must pass attestation-api verification under their selected platform policy.
- Verified platform must match the selected platform policy.
- Each local platform policy explicitly lists permitted peer platforms.
- Native raw SNP RA-TLS reports are normalized before online verification.
- Debug evidence is rejected. TDX minimum-TCB configuration is rejected until it can be enforced correctly.

## Not included

This does not yet change install, Helm/Kata runtime selection, CDS/mesh policy, inventory, scheduling, image publication, or add two-host hardware end-to-end coverage. Those are later slices required for full mixed-platform cluster operation.

## Verification

- `make test`
- `make policy-test`
- `make check-crd-chart`
- `make fmt`
- `golangci-lint run ./...`
- `go build ./...`
- targeted Linux tests, vet, diff check, and macOS arm64 cross-compile

The focused tests cover policy parsing, immutable-file enforcement, TDX/SNP selection and peer dispatch, raw SNP evidence, unknown platforms/fields, platform mismatch, invalid signatures, and REPORTDATA mismatch.


CluEleSsUK:
out of date, killing

PR #296, opened 2026-08-13
feat(node-guest-image): SEV-SNP node image — explicit platform, rendered by sync

joaosa:
## Problem

The node image only built for TDX — and not just in the build flags: `--platform tdx` was hardcoded, and **three** TDX values were baked into the image content (cred-release's platform, the NRI floor's `platform: "tdx"`, and the `confidential.ai/tdx=true` node self-label).

## Fix

`C8S_PLATFORM` is required (`tdx`|`snp`, **no default anywhere**), and one mechanism delivers every platform-dependent bit: `mkosi.sync` (which already renders the NRI floor) receives it via `--sync-input platform=…` — refusing to run without a valid value — and renders the **entire** tdx/snp divergence, three files:

- cred-release's `Environment=CRED_PLATFORM` drop-in. The unit runs `--platform=${CRED_PLATFORM}` (`=`-form: unset expands to an empty value that platform validation rejects — boot-time defense in depth behind the build-time gate). No quote-device `DeviceAllow`: cred-release opens no quote device (quotes come from attestation-api over HTTP; the operator-key measurement read is sysfs).
- the node's self-label — `confidential.ai/tdx` / **`confidential.ai/sev-snp`** (the key `c8s install --hardware-platform` actually selects on; the label is the one consumer with its own vocabulary — cred-release and the NRI floor normalize `snp` themselves).
- the NRI floor's `platform:` (a wrong value there makes CDS deny every adopted workload a certificate, hence the same no-default rule).

`confos build` gets `--platform "$C8S_PLATFORM"` for the measurement artifacts. No kernel change (confos `required.config` carries both TEEs' symbols) and no package delta (attestation-api probes both quote devices, `platforms = ["snp","tdx"]`). Since the three rendered files are the whole divergence, a one-image-serves-both design (boot-time probe + confos `--platform both`) stays evaluable later.

## SNP status — stated plainly

An SNP image **boots and attests**, but the **operator credential-release flow is TDX-only today**: the operator-key binding reads TDX RTMR[3], and SEV-SNP has no runtime-extend equivalent (`pkg/runtimemeasure`); the confos initrd also rejects an opkeydata disk on SNP. On SNP, cred-release fails closed. A working SNP operator flow needs its own binding design (HOSTDATA/ID_BLOCK or vTPM) — out of scope here.

## Measurement / CI

Rolls the TDX image measurement once (unit + rke2 config bytes) — riding the v0.4.1 reference re-seed already queued. `module-signing.crt` joins the kernel cache key (it's a kernel-fingerprint input). CI pins `C8S_PLATFORM=tdx`; SNP joins the matrix when wanted (note: an SNP build additionally needs SNP firmware at confos's `--firmware` path).

History note: rebased onto main as a single commit — earlier revisions (override drop-in, then symmetric `tdx/`/`snp/` profile-dir halves) were successively simplified away during review; the sync render replaced them all.

### DIFF
--- a/.github/workflows/c8s-image.yml
+++ b/.github/workflows/c8s-image.yml
@@ -152,7 +152,7 @@ jobs:
         uses: actions/cache@v5
         with:
           path: confidential-os-builder/output/kernel
-          key: ${{ runner.os }}-kernel-${{ env.VARIANT }}-${{ hashFiles('confidential-os-builder/mkosi/kernel-builder/mkosi.conf','c8s/node-guest-image/kernel/c8s.config','c8s/node-guest-image/kernel/c8s-dev.config','confidential-os-builder/kernel/required.config','confidential-os-builder/kernel/hardening.config','confidential-os-builder/kernel/confidential.config','confidential-os-builder/kernel/config-x86_64.snapshot','confidential-os-builder/kernel/version','c8s/node-guest-image/build') }}
+          key: ${{ runner.os }}-kernel-${{ env.VARIANT }}-${{ hashFiles('confidential-os-builder/mkosi/kernel-builder/mkosi.conf','c8s/node-guest-image/kernel/c8s.config','c8s/node-guest-image/kernel/c8s-dev.config','confidential-os-builder/kernel/required.config','confidential-os-builder/kernel/hardening.config','confidential-os-builder/kernel/confidential.config','confidential-os-builder/kernel/config-x86_64.snapshot','confidential-os-builder/kernel/version','c8s/node-guest-image/build','c8s/node-guest-image/module-signing.crt') }}
 
       - name: Cache base image tools
         uses: actions/cache@v5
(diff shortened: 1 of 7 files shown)


AmeanAsad (approved):

node-guest-image/snp/mkosi.conf
CluEleSsUK, line 1: is this meant to be all comments?

node-guest-image/tdx/mkosi.conf
CluEleSsUK, line 1: is this meant to be all comments?

CluEleSsUK (approved):

node-guest-image/snp/mkosi.conf
joaosa: yeah we don't really need this

node-guest-image/tdx/mkosi.conf
joaosa: same. I am redesigning here

Issue #264, opened 2026-08-07
Move the c8s node-image definition into this repo; consume confos as a pinned tool (kata-guest-base pattern)


joaosa:
**DONE.** The migration is complete: the node image is defined in this repo, confos is consumed as a pinned tool, reproducibility is gated in CI, and the reference-value re-seed is on its normal cadence. The one parked nice-to-have moved to #398.

## Why

The node image is a c8s artifact in everything but file location: its contents, its measurement lifecycle, and its CI (`c8s-image.yml`, since #136) live here, but the definition sits in confidential-os-builder (`mkosi.profiles/c8s/`, `bin/build-c8s`, `kernel/c8s.config`). kata-guest-base already has the right shape — definition in this repo, confos consumed as a pinned tool — and the node image should match it.

The split costs us on every change. In the first week of August alone: confos [#73](https://github.com/confidential-dot-ai/confidential-os-builder/pull/73)–[#78](https://github.com/confidential-dot-ai/confidential-os-builder/pull/78) were mostly c8s-profile fixes, and each one needed a confos merge *plus* a `CONFOS_REF` re-pin here (four re-pins that week). The `C8S_REF` default lives in a confos file but names a c8s commit, so it went stale the day after we pinned it. And the v0.3.0 changelog needed hand-scoping to keep c8s image items out of what should be a pure tool release.

## How

- confos grew two flags (confidential-dot-ai/confidential-os-builder#81, merged): `--profile-dir` stages an out-of-tree profile as if it were in-tree, and `--sync-input` hands values (like `c8s-ref`) to a profile's sync hook. `--kernel-config-fragment` already took external paths.
- `node-guest-image/` in this repo (#266) carries a byte-identical copy of the profile and kernel fragments, plus a drop-in replacement for `bin/build-c8s`. Two workflows guard the transition: a temporary byte-diff against confos at the pinned `CONFOS_REF`, and a permanent lint carrying the four invariants that used to live in confos `bin/lint` (fragment supersets, the NRI floor template rule, the nested pod-CIDR match).
- confidential-dot-ai/confidential-os-builder#84 deletes the confos-side copy (merged).

## Steps (as they actually happened)

- [x] Merge #266
[...]

PR #300, opened 2026-08-13
chore(pins): C8S_REF to v0.1.0-rc2, kata CONFOS_REF to v0.4.1

joaosa:
Two queued re-pins from #264's follow-ups:

- **`C8S_REF` → `v0.1.0-rc2`** (release tag): contains the #177-era governance work (allowlist/argv-denial/measurement-policy all in its history), and docker.yml publishes component images under `v*` tags, so the tag is directly resolvable and self-describing. Pinning the release rather than a main snapshot matches the pin's original intent — bake the component versions the deployed CLI ships as.
- **kata `CONFOS_REF` → v0.4.1** (`f7d03b7`): picks up the reproducible-kernel work. `container.config` enables no module signing, so the new builder invariant is inert for kata; **the kata kernel measurement rolls once** (seed pins), pairing with the queued reference-value re-seed.

Node-image `CONFOS_REF` is already at v0.4.1.

### DIFF
--- a/.github/workflows/kata-guest-base.yml
+++ b/.github/workflows/kata-guest-base.yml
@@ -86,7 +86,7 @@ env:
   # new pin means a new guest kernel and therefore a new SNP launch
   # measurement, and the committed kernel/config-x86_64.snapshot lockfile must
   # be refreshed in the same PR (kernel-snapshot.yml produces it).
-  CONFOS_REF: "3e6f858f93e1b85c6d1473aa4a031fdefb387710" # v0.3.0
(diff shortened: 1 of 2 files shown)


AmeanAsad (approved):

joaosa:
Snapshot-lockfile check for the CONFOS_REF bump: dispatched [kernel-snapshot.yml at f7d03b74](https://github.com/confidential-dot-ai/c8s/actions/runs/31753820082) (v0.4.1) — the produced `config-x86_64.snapshot` is **byte-identical** to the committed one, so no lockfile refresh belongs in this PR. The resolved config is unchanged between the pins (same Linux 6.16.12); what changes once is the kernel bytes/SNP measurement, from the now-pinned RANDSTRUCT/latent_entropy seeds. The kata fragment enables no module signing, so the new v0.4.1 MODULE_SIG_KEY builder invariant is safe.

PR #315, opened 2026-08-14
fix(kata-policy): admit only guest-local storages, deny diagnostic reads

CluEleSsUK:
The baked policy placed no driver, filesystem or mount-point constraint on the storages a CreateContainerRequest carries. The host could therefore register host-backed storage at a path the runtime manages and bind it over an injected sidecar's /run/c8s/certs or /run/c8s/secrets, reading the pod's mesh leaf private key and every released secret in plaintext while the release predicate was satisfied genuinely.

A storage is now admitted as the guest pull, as an ephemeral tmpfs, or as a blk/blk-ccw/scsi device carrying encryption_key=ephemeral with a non-absolute source, and its mount point must sit one plain component past one of the two trees the runtime manages. Driver and fstype are both pinned because either alone is forgeable: the ephemeral handler mounts whatever filesystem it is handed, so fstype decides whose bytes appear, while watchable-bind copies its source into its mount point on a timer without calling mount(2) at all, and the overlay handler rewrites fstype to overlay on one of its options. The marker rather than the driver gates the block shape because it is what routes the device to CDH, which LUKS2-formats it under a key the guest generates; an absolute source there names a device node the block handler resolves directly, and this branch formats what it resolves.

CreateSandboxRequest and UpdateEphemeralMountsRequest carry the same constraints, and additionally reject the multi-layer and overlay option forms. is_multi_layer_storage dispatches on those options before the driver is consulted, so pinning the driver alone leaves an overlay assembled out of host block devices at the CopyFile seeding tree, which is the tree every container binds from. The encrypted-block shape is a rule of its own, reached only from the container path: the sandbox source tokens are all non-absolute, so a marked block device was admissible there until it was separated out.

GetDiagnosticDataRequest is denied. Its termination-log branch reads the guest file behind the OCI mount whose destination equals the terminationMessagePath annotation and returns four kilobytes of it over vsock; both fields are host-written, so it reads any guest path a mount may name, and nothing in c8s consumes it. The general point is worth recording: a policy default of true is an unaudited grant whatever the RPC is called, and this one was missed at first because only the RPCs whose names implied I/O were examined.

Non-bind OCI mounts are held to the pseudo-filesystems, since rustjail passes the declared type straight to mount(2) and the guest kernel carries virtiofs, 9p and the disk filesystems. The source-prefix branch now requires an actual bind and the same component-past-the-tree shape a storage mount point has: a virtiofs or 9p source is the tag the host gave the device rather than a path, so it can be spelled as one of the sandbox directories.

Block-backed PVCs are denied as a consequence, as are default-medium emptyDirs and hugepages outside the block-encrypted shim mode. Under the shipped emptydir_mode="block-encrypted" the admitted set covers every shape the runtime produces; widening it for another mode costs a guest image rebuild and an operator pin bump.


kata-guest-base/extra/etc/kata-opa/default-policy.rego
CluEleSsUK, line 1: I've gotta be honest, these .rego files are a bit mysterious to me

PR #369, opened 2026-08-18
fix(policy-monitor): make the in-guest inventory and init-data pin work on TDX

AnomalRoil:
Two guest-side fixes without which `--cvm-mode=pod --hardware-platform=tdx` cannot run any `confidential.ai/cw` workload. Both found bringing up pod-as-CVM on a bare-metal TDX host (Xeon 6527P, RKE2); the tdx metal e2e lane runs `--cvm-mode node` and never exercises the in-guest policy-monitor.

**1. Sandbox-digests endpoint stamped SEV-SNP on every platform.** `startSandboxDigests` hardcoded `PlatformSnp`, so on a TDX guest the served leaf carried a TDX evidence envelope under the SNP TEE type and CDS refused every sandbox token:

```
csr_denied: resolve inventory key: workloadclaims: reach inventory 10.42.0.83: Get "https://10.42.0.83:1019/identity": ratls: peer attestation failed: ratls: unsupported TEE platform: online verification not implemented for platform "tdx"
```

CDS and tls-lb never hit this because their platform comes from chart values; the guest-side inventory had no such input. Now read once from the in-guest attestation-api (`platform: auto`, the same call every RA-TLS handshake makes) — the source `RATLSEvidence` already dispatches on.

**2. Init-data pin refused on TDX.** The shim commits `sha256(document)` into MRCONFIGID zero-padded to 48 bytes (the shape attestation-rs's TDX verifier binds `expected_init_data_hash` against), but policy-monitor only accepted SNP's 32-byte HOST_DATA, logged `not a 32-byte anchor`, never learned the CDS pin, and enforced the baked seed alone — so under pod-mode TDX no `c8s allowlist add` ever reached a running guest and any image outside the baked seed was refused. Accept a zero-tailed 48-byte claim and take the 32-byte prefix; a non-zero tail stays ill-shaped. `docs/kata-image-policy.md` updated (it documented TDX as having no path).

Both are baked into kata-guest-base, so they need a guest-image rebuild to take effect. Validated live on the TDX host with a guest image built from this branch.

### DIFF
--- a/docs/kata-image-policy.md
+++ b/docs/kata-image-policy.md
@@ -305,11 +305,12 @@ field rather than passing silently.
 On a mismatch, or with no document, the guest falls back to the baked
 seed.
 
-**TDX has no equivalent path yet:** the digest goes to `MRCONFIGID`,
-which is 48 bytes where the anchor is 32, so the guest refuses the
-claim and enforces the baked seed alone. Empty `cds.measurements` (the chart
-default) also leaves the refresh disabled on every platform — operator
-additions then reach the host-side enforcer but not running guests.
+**TDX takes the same path through `MRCONFIGID`:** the shim commits
+`sha256(document)` zero-padded to the 48-byte `MRCONFIGID`, and the guest
+accepts that shape as the 32-byte anchor (a non-zero tail is refused as
+ill-shaped). Empty `cds.measurements` (the chart default) leaves the refresh
+disabled on every platform — operator additions then reach the host-side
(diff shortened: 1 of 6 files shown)


joaosa (approved):

CluEleSsUK (approved):

AnomalRoil:
That this wasn't caught before now shows we really need CI to test both TDX and SEV-SNP in all modes we are supporting. cc @b0xtch

PR #455, opened 2026-08-20
security(tdx): pin RTMR[1]/[2] on the in-cluster verification paths

AnomalRoil:
On TDX the launch digest (MRTD) covers the TDVF firmware alone; the guest kernel measures into RTMR[1] and the command line — which carries the dm-verity root hash of the rootfs — into RTMR[2]. `ratls.VerifyPolicy.RTMRs` and its enforcement (`attestationclient`) existed, but only the operator CLIs ever populated it: no in-cluster verifier pinned registers, so on TDX the in-cluster measurement allowlist proved only "some TD booting this firmware".

This threads RTMR pins through every in-cluster verification path:

- **Shared plumbing**: `ratls.Pins` (measurements + RTMRs) and `ratls.ParseRTMRPins` / `ParseRTMRPinsString` (`<index>=<sha384-hex>`, indices 1–3; RTMR[0] refused since it varies with the pod's vCPU/memory shape). `NewVerifyingHTTPClient` now takes `Pins`.
- **CDS issuance**: `--rtmrs` gates `/attest` and `/attest-key` on TDX-shaped evidence via the exported `attestationclient.EnforceRTMRs`; a pinned register the report does not carry is a refusal. `--handoff-rtmrs` pins the CA-adopt dial; the sandbox-digests callback pins both directions.
- **ratls-mesh**: `--rtmrs` / `--cds-rtmrs` (host mode), `C8S_MESH_RTMRS` / `C8S_CDS_RTMRS` (in-guest).
- **Everything dialing CDS**: get-cert, get-secret, get-volume, nri-image-policy (`cds_rtmrs`), policy-monitor (flag, env, and the launch-committed init-data key `c8s.cds.rtmrs` — delivered and digest-checked exactly like the measurements), allowlist-proxy, request-handoff.
- **Chart / operator**: `cds.rtmrs` and `ratlsMesh.rtmrs` fan out to all of the above, mirroring the dual use of `cds.measurements`; the webhook emits the pins into injected sidecar args and the kata init-data document.
- **install / render-values**: `--rtmrs` emits both value paths; a `--hardware-platform=tdx` install without RTMR pins now warns that measurement pinning covers firmware only.

SNP evidence is unaffected throughout: `kernel-hashes` folds the guest image into its launch digest, so the pins are enforced only against TDX-shaped evidence and a mixed fleet keeps working.

Limitations, stated in the docs: the in-cluster pin is one register set for the fleet (not per-image tuples like `c8s verify --image-manifest`), and RTMR[1]/[2] values are still not derivable offline — they come from a boot trusted for other reasons (`c8s verify` verbose output). `docs/ratls.md` and `docs/kata-image-policy.md` updated accordingly.


joaosa (approved):
nice! I had been thinking about how this was missing

PR #464, opened 2026-08-21
fix(install): scope the cw DNS carve-out to this cluster's resolver

CluEleSsUK:
`ratlsMesh.clusterDNSIP` defaults to `10.53.0.10`, the c8s node image's `cluster-dns`. It reaches `iptables-sync` as `--cluster-dns-ip`, which destination-scopes the UDP/53 carve-out in a chain that otherwise drops every non-TCP packet a `confidential.ai/cw` pod sends. On any cluster whose CoreDNS Service sits elsewhere — stock RKE2 `10.43.0.10`, kubeadm and kind `10.96.0.10` — the carve-out names an address no pod resolves against and every cw pod loses DNS, while `c8s install` exits 0 reporting a healthy release.

## Why it was undiagnosable

The symptom points away from the cause:

```
lookup c8s-cds.c8s-system.svc on 10.43.0.10:53: i/o timeout
```

`10.43.0.10` is the correct server. The query is dropped because the carve-out names `10.53.0.10`, so the natural reading is that CoreDNS is unhealthy — it isn't. Blast radius isn't limited to volumes: `get-cert` and anything else resolving a Service name fails the same way.

The mismatch was already diagnosed one layer too late. `warnUnlessClusterDNSResolves` (`pod_ipsets_linux.go`) compares the configured address against the kubelet-written `/etc/resolv.conf` the hostNetwork mesh DaemonSet shares with every pod on the node, and `TestWarnUnlessClusterDNSResolves` already asserts exactly this configuration. It surfaces as a WARN inside a DaemonSet that is otherwise healthy.

## Fix

`c8s install` reads the ClusterIP off the kube-system Service labelled `k8s-app=kube-dns` and renders it — beside the cluster facts the install already derives (`clusterDistroNodes` reads kubelet versions, `inventory_cidr.go` reads `spec.podCIDRs`).

The label, not the name, is the portable handle: distributions rename the Service (RKE2 calls it `rke2-coredns-rke2-coredns`) but all carry the label. That name-guessing is exactly what `c8s.tlsLb.resolver` has to do off the distro today.

Zero matches, or several on different ClusterIPs, fails the install before anything is mutated. A value the guard cannot use has no legitimate use, and picking one of several would let anyone able to create a kube-system Service *choose* the address rather than merely deny it. The error names `ratlsMesh.clusterDNSIP` and a pod's own `resolv.conf` as the authority, which is also the answer where the two disagree — under a node-local DNS cache pods query its link-local address, not the Service.

## Precedence and scope

- Derivation runs in `installCmd.RunE`, not `buildValueArgs`, which `c8s render-values` shares and which never touches a cluster (now pinned by a test).
- The computed values are helm's last `-f`, so a `-f` file that sets `ratlsMesh.clusterDNSIP` owns the address and the lookup stands aside — the rule `valuesFilesSetDistro` applies to the containerd layout.
- Under `--cvm-mode=pod` the chart renders `ratlsMesh.enabled=false` and the guest guard reads `C8S_CLUSTER_DNS_IP`, so nothing is read and nothing is rendered.
- No new CLI flag: `ratlsMesh.clusterDNSIP` is already the documented override and the error names it.
- The chart value stays scalar, so a dual-stack resolver contributes its primary address and the install says on stderr which family goes unguarded.

## Tests

`cluster_dns_test.go` covers `clusterDNSServiceIP` (RKE2's renamed Service, a headless sibling skipped, zero matches, two Services on different ClusterIPs, an unparseable ClusterIP, malformed JSON, kubectl's own stderr surfaced, dual-stack primary) and `resolveClusterDNSIP` (host mode derives, pod mode reads nothing, a `-f` that sets the key stands the lookup down, a `-f` that doesn't still derives, an underivable cluster fails).

`TestInstallScopesTheDNSCarveOutToThisCluster` runs the CLI end to end and asserts the lookup precedes both `kubectl apply` and `helm upgrade` and that `ratlsMesh.clusterDNSIP: 10.43.0.10` — not the chart default — lands in the computed values. `TestInstallAbortsWhenTheClusterDNSIsUnreadable` asserts neither runs when the lookup fails. `TestInstallKeepsAnOperatorsClusterDNSIP` asserts no `kubectl get svc` and no computed key. All three were mutation-checked: dropping the emit or the `-f` gate fails them.

`TestChartIptablesSyncCarriesClusterDNS` gains an override render, so the last hop is covered — until now a hardcoded `10.53.0.10` in the DaemonSet would have passed.

The integration lane drops the lookup it was doing by hand (`run.sh` did it with the comment *"Read it off the cluster, as an operator must"*). kind's `10.96.0.10` against the `10.53.0.10` default now makes the lane a real regression test, proven end to end by the cw pod's `nslookup` it already runs; a direct assertion on the DaemonSet argv sits right after the install so a broken derivation names itself.

The mesh e2e keeps its independent check but selects on the label instead of the name `kube-dns`, which does not exist on RKE2 — the assertion was wrapped in `if [ -n "$cluster_dns" ]` and therefore silently no-opped on the one distribution this fixes.

## Two things deliberately not in this PR

**The guest side.** `docs/ratls.md` says the guest reads `C8S_CLUSTER_DNS_IP` "baked in kata-guest-base". Nothing in the tree writes that variable — not `cloudinit.env`, not the chart, not the pod mutator — so under `--cvm-mode=pod` guests always fall back to the compiled-in `10.53.0.10` with no supported override. It also can't simply be baked per-cluster, since that file lives in the dm-verity-measured rootfs. Left alone here; worth its own issue.

**Deriving inside the DaemonSet instead.** The mesh pod is `hostNetwork` + `ClusterFirstWithHostNet`, so its `/etc/resolv.conf` *is* the answer — it handles node-local DNS caches and dual-stack lists that a Service lookup cannot, and it would fix chart-only and `helm upgrade --reuse-values` installs too. It is also a smaller change. I did not take it because `pod_ipsets_linux.go` states the opposite as an invariant — "resolv.conf is read to contradict a wrong value, never to supply one, so a resolver the node was pointed at cannot widen a fail-closed egress guard" — and inverting that is a trust-boundary decision, not a PR-sized one. Worth deciding explicitly; the two are not mutually exclusive.


CluEleSsUK:
closing this to rethink how we do DNS

PR #494, opened 2026-08-25
feat(volumes): mutable encrypted volumes alongside immutable ones

CluEleSsUK:
c8s volumes have been immutable after creation: erofs under dm-verity inside plain dm-crypt. Some workloads need to write — scratch space, datasets they update, checkpoints. This adds a mutable mode that keeps the same encryption and release path and drops the verification layer, which cannot cover a writable device.

`c8s volume create --mutable` builds a writable ext4 of --size bytes (preloaded from --source when given; --size inferred from the source otherwise, or required without one). The default stays immutable.

- The key blob carries the mode: "mutable": true with no verity geometry, and Validate refuses a blob with both or neither. Old escrow blobs decode as immutable; old daemons reject the new field and fail closed.
- volumed reads the mode from the validated blob: writable dm-crypt, no verity mapping, ext4 mounted rw. Immutable is unchanged. The open-idempotency commitment covers key+mode+root hash, so a mode flip on an open volume is not authorized.
- One writer per device: two rw mounts of one device corrupt the filesystem (a rolling update hits this on a single-device workload), so a device with a live mount refuses a mutable open, and one with a mutable mount refuses every open — 409 at the daemon.
- mkfs.ext4 -e remount-ro: an fs error degrades the mount read-only rather than spreading corruption; on a no-integrity volume the likely cause is the host flipping bits.
- The webhook no longer pins the container mount read-only (writability is the daemon's mount flags) and the kata qemu wrapper attaches devices read-write — the host cannot know the mode, since it lives in the key blob it never sees; immutable enforcement is the guest's dm-crypt --readonly either way.
- The uninstall teardown hook unmounts c8s-crypt-* sources too, not only c8s-verity-*.

Accepted and documented: a mutable volume has no integrity — bit flips and rollbacks are undetected. Per-sector authentication (AEAD) is future work. No resize, re-encryption, or mode conversion.

### DIFF
--- a/docs/volumes.md
+++ b/docs/volumes.md
@@ -69,9 +69,11 @@ anchors nothing. It is also a fingerprint naming which model a volume holds, and
 A mutable blob carries the key and its mode, nothing else:
 
 ```json
-{ "type": "c8s.volume/v1",
+{
+  "type": "c8s.volume/v1",
   "key": "<base64, 64 bytes>",
-  "mutable": true }
+  "mutable": true
+}
 ```
 
 There is no integrity anchor to carry — a volume the host can rewrite has
@@ -105,7 +107,8 @@ c8s volume create --mutable \
   --path /tenant-a/volumes/datasets \
   --escrow-out ./datasets.escrow.json \
   --node node-1 \
-  --url https://cds.example --measurements-file ./m.txt \
+  --url https://cds.example \
+  --measurements-file ./m.txt \
   --operator-key ./operator.key
 ```
 
@@ -113,10 +116,14 @@ or empty, for scratch space the workload fills itself:
 
 ```sh
 c8s volume create --mutable \
-  --name scratch --size 50Gi \
-  --out ./scratch.img --path /tenant-a/volumes/scratch \
-  --escrow-out ./scratch.escrow.json --node node-1 \
-  --url https://cds.example --measurements-file ./m.txt \
+  --name scratch \
+  --size 50Gi \
+  --out ./scratch.img \
+  --path /tenant-a/volumes/scratch \
+  --escrow-out ./scratch.escrow.json \
+  --node node-1 \
+  --url https://cds.example \
+  --measurements-file ./m.txt \
   --operator-key ./operator.key
 ```
 
@@ -237,7 +244,9 @@ Each entry is `NAME=/store/path`. `NAME` selects the node's device by its
 the volume dir — above, `/models/weights`. It must be a DNS-1123 label of at
 most 12 characters, because the serial holds no more. The annotation is the
 same for both modes; whether the plaintext is writable comes from the volume's
-key blob, not from the pod.
+key blob, not from the pod. Flipping the mode in a blob gets nothing: an
+immutable image is erofs and fails to mount as ext4, and a mutable image has
+no hash tree for verity to anchor.
 
 The webhook injects a `c8s-volume` sidecar and, per volume, an `emptyDir` with
 `mountPropagation: HostToContainer`. The sidecar fetches the key over the
@@ -417,6 +426,9 @@ of scope by decision, not by oversight.
   `c8s cds verify --operator-keys`, and running it continuously is a
   precondition for trusting a volume.
 - **Access patterns are visible.** Which sectors are read, and when, leaks
-  structure.
+  structure. AES-XTS is deterministic, so for a mutable volume the host can
+  also tell when a sector returns to a value it has seen before — the standard
+  FDE trade-off (BitLocker and FileVault share it), with no IV space per
+  sector write to do better with.
 - **Availability.** The host can withhold or destroy the device.
 - **Whatever the workload does with the plaintext** once it has it.
--- a/docs/volumes.md
+++ b/docs/volumes.md
@@ -20,13 +20,17 @@ already has. Treat volume keys accordingly.
 
 ## The artifact
 
-An image built by `c8s volume create`, in three layers:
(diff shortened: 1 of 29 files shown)


docs/volumes.md
```diff
@@ -62,8 +66,22 @@ entry because it is the integrity anchor, and an anchor the host can edit
 anchors nothing. It is also a fingerprint naming which model a volume holds, and
 `GET /allowlist` is unauthenticated.
 
+A mutable blob carries the key and its mode, nothing else:
+
+```json
+{ "type": "c8s.volume/v1",
+  "key": "<base64, 64 bytes>",
+  "mutable": true }
```
CluEleSsUK:
```suggestion
{ 
  "type": "c8s.volume/v1",
  "key": "<base64, 64 bytes>",
  "mutable": true 
}
```

docs/volumes.md
```diff
@@ -76,9 +94,42 @@ c8s volume create \
...
+```sh
+c8s volume create --mutable \
+  --name datasets \
+  --source ./imagenet \
+  --size 200Gi \
+  --out ./datasets.img \
+  --path /tenant-a/volumes/datasets \
+  --escrow-out ./datasets.escrow.json \
+  --node node-1 \
+  --url https://cds.example --measurements-file ./m.txt \
```
CluEleSsUK:
```suggestion
  --url https://cds.example \
  --measurements-file ./m.txt \
```

docs/volumes.md
```diff
@@ -76,9 +94,42 @@ c8s volume create \
...
+  --operator-key ./operator.key
+```
+
+or empty, for scratch space the workload fills itself:
+
+```sh
+c8s volume create --mutable \
+  --name scratch --size 50Gi \
+  --out ./scratch.img --path /tenant-a/volumes/scratch \
```
CluEleSsUK:
```suggestion
  --out ./scratch.img \
  --path /tenant-a/volumes/scratch \
```

docs/volumes.md
```diff
@@ -76,9 +94,42 @@ c8s volume create \
...
+  --escrow-out ./scratch.escrow.json --node node-1 \
```
CluEleSsUK:
```suggestion
  --escrow-out ./scratch.escrow.json \
  --node node-1 \
```

docs/volumes.md
```diff
@@ -76,9 +94,42 @@ c8s volume create \
...
+  --url https://cds.example --measurements-file ./m.txt \
```
CluEleSsUK:
```suggestion
  --url https://cds.example \
  --measurements-file ./m.txt \
```

docs/volumes.md
```diff
@@ -76,9 +94,42 @@ c8s volume create \
...
+  --operator-key ./operator.key
+```
+
+`--size` takes a byte count or a quantity like `50Gi`. Omitted with a
+`--source`, it is inferred — the tree's bytes plus a block per entry, grown by
+half — and printed, so you can see what was chosen and pin it on the next run.
+It is rejected without `--mutable`: an immutable image sizes itself to its
+contents. Every byte of the image is encrypted at creation, free space
+included, so a `--size 200Gi` build writes a 200Gi file.
```
CluEleSsUK: is this CPA-secure using AES-XTS?

docs/volumes.md
```diff
@@ -162,8 +218,8 @@ Release is gated on the workload entry's `secrets` grant, exactly as in
 volume beneath it, and the annotation naming which volume to open is
 host-written. `create` prints an exact-path grant for this reason.
 
-`read` only. A volume is mounted read-only, so a write grant says nothing about
-whether a workload may see the plaintext.
+`read` only. Writability is a property of the volume, not the grant: a write
```
CluEleSsUK: what happens if we try and mount one of the readonly volumes as mutable and write to it?

internal/cmds/volume/image.go
```diff
@@ -130,6 +139,158 @@ func Build(ctx context.Context, cfg BuildConfig) (Verity, error) {
...
+		}
+		return nil
+	})
+	if err != nil {
+		return 0, 0, fmt.Errorf("volume: measure --source: %w", err)
+	}
+	return dataBytes, entries, nil
+}
+
+func alignUp(n, unit uint64) uint64 { return (n + unit - 1) / unit * unit }
```
CluEleSsUK:
```suggestion
func alignUp(n, unit uint64) uint64 { 
    return (n + unit - 1) / unit * unit 
}
```

internal/cmds/volumed/open.go
```diff
@@ -139,23 +153,39 @@ func (o *Opener) Open(ctx context.Context, req Request) error {
...
 	o.mu.Lock()
 	existing, lost := o.mounts[mountKey(req.Pod.UID, req.Name)]
-	if !lost {
+	conflict := !lost && o.deviceConflict(req.Device, req.Blob.Mutable)
+	if !lost && !conflict {
 		o.mounts[mountKey(req.Pod.UID, req.Name)] = m
 	}
 	o.mu.Unlock()
-	if !lost {
+	if !lost && !conflict {
```
CluEleSsUK: this is a code smell to do the same check twice within a few lines

internal/cmds/volumed/open.go
CluEleSsUK: in general there is a lot of branching in this file: it might make sense to hide the mutable and immutable versions behind an interface and just implement in two different structs for ease of reading

PR #503, opened 2026-08-27
feat(cds)!: remove the attested mesh-CA handoff

CluEleSsUK:
/handoff is unworkable by design: the operator must know the replacement CDS's launch measurements upfront — the very thing a roll is supposed to avoid — and a weak base measurement set lets a malicious node pass the gate and extract the mesh CA private key. Remove the feature rather than patch a broken trust story:

- drop POST /handoff and the issuer handoff protocol (EAR-gated wrap/unwrap, self-provisioned signer EAR bootstrap, expiry metric)
- drop pull-on-startup CA adoption; CDS always generates a fresh in-memory mesh CA
- drop the `c8s cds request-handoff` probe CLI
- drop --handoff-measurements/-peer-url/-peer-timeout/-rtmrs and the cds.handoff.{enabled,peerUrl} chart values; stale values files now fail schema validation
- drop the operator_keys_hash binding from /attest-key (handoff was its only verifier); /attest-key itself stays for the /sign-csr EAR flow, with an unchanged REPORTDATA wire format for hash-free requests
- drop allowlist Store.RestoreSnapshot (adoption was its only caller), resources.Handoff, ratls.ReportDataForKeyWithContext, and the ca-handoff e2e script, Makefile target, and snp-metal CI suite

Behavior changes: a CDS restart always re-bootstraps the mesh CA (the documented singleton posture); /secrets is no longer gated off when handoff was configured; the cds Deployment always rolls Recreate.


CluEleSsUK:
merging this as the tls-lb vuln error is being suppressed in this PR: https://github.com/confidential-dot-ai/c8s/pull/506

PR #511, opened 2026-08-31
getcert: exit once the installed leaf expired and renewals keep failing

AnomalRoil:
A tls-lb served an expired mesh leaf for five days: the c8s-cert renewal loop retries forever, exec liveness probes are policy-denied on locked guests (`ExecProcessRequest := false`), and nginx keeps serving whatever certificate it has — so the failure was silent and sticky, with `attest-pq` answering 503 and every pinned client failing closed. A pod restart recovered it immediately.

Two changes:

- **getcert**: once the installed leaf is past `NotAfter` and `expiredExitFailures` (3) consecutive renewals have failed, the loop exits instead of retrying. As a native sidecar (`restartPolicy: Always`) the container restarts with fresh client state and re-runs the full issuance — the only self-heal available where exec probes are blocked. A still-valid leaf keeps the existing retry/backoff behavior.
- **cds**: route `net/http` server error lines through a filter that demotes the TLS-handshake EOF/reset lines kubelet's tcpSocket probes generate (one per probe, every few seconds) to debug, so real handshake faults stay visible.

The root cause of the renewal attempts failing on the incident cluster is still unknown (logs rotated); a watch is running to catch a recurrence with fresh logs.

### DIFF
--- a/docs/ratls.md
+++ b/docs/ratls.md
@@ -681,6 +681,11 @@ Poll timing never changes the match decision.
 
 Every delay is also capped at half the installed leaf's remaining lifetime, and
 a failed renewal retries on a short backoff rather than after a full interval.
+Once the installed leaf has **expired** and renewals still fail, get-cert exits
+instead of retrying forever: as a native sidecar it restarts with fresh client
+state and re-runs the full issuance. A locked guest denies exec probes, so a
+process that keeps running while serving a dead certificate would otherwise
+never be restarted by anything.
 The named-leaf TTL is the shortest CDS issues and `certutil` does not backdate
 `NotBefore`, so a renewal interval alone — the chart's `renewInterval` — is not
 a safe schedule: it must stay strictly below `cds.namedCertTTL`, and the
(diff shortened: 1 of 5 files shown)


CluEleSsUK:
`get-cert` supports a `--renew-interval` - shouldn't it get renewed before expiry?

AnomalRoil:
@CluEleSsUK it should yeah afaict and yet didn't and when it didn't it failed very silently. This fixes that, hopefully

CluEleSsUK:
it fixes it in the sense that certs now get refreshed, though the `--renew-interval` flag isn't used here, so this is more of a workaround. We should either remove it and rely on bouncing the pod or fix its use

AnomalRoil:
I'm not sure what you're getting at here, but what this does is adding a hard limit on the number of failures we accept in the renew loop before actually failing for real. 

Previously we could get stuck in a renew loop in some weird cases I couldn't reproduce but did hit on a test cluster without ever properly failing. 
`renew-interval` is used by `renewalInterval` in the renewLoop, isn't it?

PR #516, opened 2026-09-01
fix(getsecret,getcert): enforce the memory-only invariant on secret outputs

joaosa:
getsecret documents that `--out-dir` must be memory-backed, and getcert writes the workload private key to a volume the webhook requires to be a Memory-medium emptyDir. Neither binary checked — on any other filesystem the secret bytes land where the host can read them.

Both now go through `fileutil.RequireRAMBacked`, copied verbatim from `feat/attested-join` so that branch still merges cleanly. getsecret refuses a non-tmpfs `--out-dir` before touching the secret store; getcert refuses a `--key-out` on persistent storage at startup, next to the existing writability probe. The cert and CA outputs are public and stay unconstrained. Off Linux the check fails closed.

The second commit pins the WriteAtomic contract these writers rely on: overwriting now provably tightens permissions instead of inheriting the old mode. The third moves the flag-prefixed wrapper into cmdsutil, next to the other shared flag validators.

This is a behavior change for anyone pointing these flags at persistent storage. The refusal is loud at startup, before any secret is fetched or key generated, and the chart/webhook path is unaffected — the webhook already rejects non-Memory volumes for both.

Tests mirror join's TestPrepareTokenDir and skip where the temp dir is tmpfs. `TMPDIR=/tmp go test` passes on the touched packages; the default macOS TMPDIR trips a pre-existing unix-socket path-length failure, reproduced on clean main. Linux cross-build clean.

### DIFF
--- a/internal/fileutil/rambacked_other.go
+++ b/internal/fileutil/rambacked_other.go
@@ -2,10 +2,22 @@
 
 package fileutil
 
-import "fmt"
+import (
+	"fmt"
+	"os"
+)
 
 // RequireRAMBacked fails closed off Linux: without statfs there is no way to
 // prove dir is memory-backed, so secret staging is refused.
 func RequireRAMBacked(dir string) error {
 	return fmt.Errorf("cannot verify %s is RAM-backed: linux only", dir)
+}
+
+// RequireRAMBackedRoot fails closed off Linux, where the filesystem behind an
+// open directory handle cannot be proven memory-backed.
+func RequireRAMBackedRoot(root *os.Root) error {
+	if root == nil {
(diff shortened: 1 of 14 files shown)


internal/fileutil/rambacked_other.go
AnomalRoil, line 1: I'm not convinced we are planning on supporting non-linux platforms, do we?

AnomalRoil (approved):
I guess you do need the distinction so that tests are passing on Darwin?

internal/fileutil/rambacked_other.go
joaosa: We aren't but this pattern ends up getting re-added. We should do a targeted cleanup on this front

PR #518, opened 2026-09-01
feat(node-guest-image): fail closed when the scratch disk is missing

joaosa:
A node VM provisioned without its `confai-scratch` disk boots on the initrd's 2G tmpfs fallback, registers Ready, and wedges under load.

- `scratch-enforce.service` + `scratch-enforce.sh`, mirroring the gpu-cc-enforce pattern (#507): oneshot before the rke2 pair, `RequiredBy=` both, `FailureAction=poweroff-force`. The gate checks what the initrd actually did — the plain-mode dm mapping named `scratch` — rather than re-scanning disks, and refuses a mapping below the 64G floor (decimal, so 64GB and 64GiB disks both pass; an undersized disk re-creates the wedge with more rope).
- Root-free unit test (`make test-node-guest-image-scratch`), wired into node-guest-image lint; shared `not`/`stderr_has` helpers moved into `tests/lib.sh`.
- Invariants: `tdx-metal-e2e.yml` must keep `serial: confai-scratch` (joined to the existing re-vendor marker loop), and the pinned confos checkout must still open the scratch disk as dm `scratch` — a confos rename would otherwise turn this gate into a poweroff of every healthy node.
- Docs: launch requirements live in `node-guest-image/README.md` (also naming the dm mapping, the enforced floor, and confidential-metal's default-on datadisk), and `mkosi.conf` points there instead of the nonexistent `docs/C8S-IMAGE.md` (c8s-workspace#178).

Provisioning-side defaults are confidential-metal#208; this gate is defense in depth and stays useful after it lands.

### DIFF
--- a/node-guest-image/README.md
+++ b/node-guest-image/README.md
@@ -46,7 +46,8 @@ Every node VM needs a write-storage disk with virtio-blk serial
 `confai-scratch`, at least 64G. The initrd encrypts it and uses it as the
 rootfs upper. Without it, the guest boots anyway on a 2G RAM tmpfs, comes
 up Ready, and wedges once RKE2 fills it — that's a flapping node, not a
-boot error. The serial is what matters: labels don't work, and the disk
+boot error. `scratch-enforce.service` closes that hole: it powers the VM
+off before rke2 starts when the initrd fell back to tmpfs. The serial is what matters: labels don't work, and the disk
 must be one of `/dev/vdb`–`vdd`. In KubeVirt that's `disk: {bus: virtio}`
 plus `serial: confai-scratch` (see the vendored `tdx-metal-e2e.yml`); with
 confidential-metal, pass `--datadisk-gi` — its default is 0, meaning no
--- a/node-guest-image/README.md
+++ b/node-guest-image/README.md
@@ -43,15 +43,17 @@ Layout:
 ## Launch requirements
 
 Every node VM needs a write-storage disk with virtio-blk serial
-`confai-scratch`, at least 64G. The initrd encrypts it and uses it as the
-rootfs upper. Without it, the guest boots anyway on a 2G RAM tmpfs, comes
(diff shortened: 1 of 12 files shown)


node-guest-image/tests/gpu-node-label-test.sh
```diff
@@ -12,8 +12,6 @@ SCRIPT=${GPU_LABEL_SCRIPT:-"$TESTS_DIR/../c8s/mkosi.extra/usr/local/bin/gpu-node
 [[ -x "$SCRIPT" ]] || { echo "script not found: $SCRIPT"; exit 2; }
 
 # negation wrapper: ok runs "$@" directly, so a bare ! is not a command.
```
AnomalRoil:
```suggestion
```

node-guest-image/README.md
```diff
@@ -40,6 +40,25 @@ Layout:
...
+(`--datadisk-gi`, 0 opts out).
+
+The initrd encrypts the disk and mounts it as the rootfs upper, via a dm
+mapping named `scratch`. Without the disk it falls back to a 2G RAM tmpfs:
+the guest comes up Ready, then wedges once RKE2 fills it — a flapping
+node, not a boot error. `scratch-enforce.service` closes that hole by
+checking for the dm mapping and powering the VM off before rke2 starts.
+
+A second disk with serial `confai-containerd` is recommended so the image
+cache stays off guest RAM. `confai-models`, `opkeydata`, and `joindata`
```
AnomalRoil: Are all these and what they are helping with documented somewhere else?

AnomalRoil (changes requested):
Just a comment to remove and lgtm

node-guest-image/README.md
joaosa: good point. moved the docs for all disk configs to the README

AnomalRoil (approved):
lgtm

PR #522, opened 2026-09-02
feat(cds): seal the allowlist into an attested mesh CA (--static-allowlist)

AnomalRoil:
Adds a sealed-policy mode so an end user can verify the operator is not swapping pods or widening policy between two HTTPS requests to a workload behind tls-lb — and roots that seal in the launch measurement, not in CDS's word.

**CDS `--static-allowlist`**: the seed document becomes the one immutable allowlist for the instance's lifetime (the store is **replaced**, every write endpoint is disabled), and the mesh CA certificate is minted carrying the standard `…1.1` RA-TLS extension over the **CA public key** plus `…1.3` = `SHA-256(Allowlist.Canonical())` (`pkg/ratls/staticallowlist.go`). attest-pq/attest-lb already commit `SHA-256(mesh_CA_DER)` into nonce-fresh `report_data`, so the sealed digest is covered by per-request hardware evidence with no protocol change.

**What roots the stamp, per shape** (the CA's evidence binds only the key to a measurement, and every pod of a node image gets that evidence):
- **Node-as-CVM**: the policy is baked into the node image (`node-guest-image/build C8S_STATIC_ALLOWLIST=<doc>`, or the image workflow's new `static_allowlist` input, published as `…-sa<digest>` tags with the document in the artifact). The baked NRI plugin enforces exactly it with no pull (`allowlist.static_path`), so the measurement admits only the sealed set on any node. `c8s install --static-allowlist --bootstrap-allowlist <doc>` mounts the baked file into CDS and pins its digest (`cds.staticAllowlistDigest`); CDS refuses to start on a mismatch (`--static-allowlist-digest`). `c8s render-allowlist` composes the document.
- **Pod-as-CVM**: the install renders the chart seed, pins its digest and launch-commits it through the CDS pod's kata init-data (`cds.initDataAnnotation`, `c8s.cds.allowlist-seed-sha256`); CDS verifies its own report's init-data claim against that document before sealing (`--static-allowlist-init-data`).

**Verifier**: `c8s verify --static-allowlist` (requires `--mesh-ca`) enforces the stamp, verifies the CA's embedded evidence against the same measurement policy as the target (plus `--init-data` on pod-as-CVM), cross-checks `--allowlist` bytes against the sealed digest, and fails a leaf whose `…1.5` stamp was decided under another policy. `c8s allowlist digest` prints the pin value. The operator-keys install guard now inspects `-f` files instead of waving any of them through.

**Baking locally**: `node-guest-image/build C8S_GPU_ATTEST=0` composes the plain `attest` profile (driver, CC enforcement and signed modules unchanged) so an operator bake needs no libnvat, attestation-rs build or `ATTEST_GPU_BIN`/`LIBNVAT`; c8s consumes no GPU evidence today. `CONFOS_REF` moves to confos `main` with [confidential-os-builder#122](https://github.com/confidential-dot-ai/confidential-os-builder/pull/122), which makes the GPU-attest staging fail closed on a host that cannot resolve libnvat's deps (no measured content changed). `node-guest-image/README.md` gains the build recipe.

Design, trust argument and residual gaps: docs/static-allowlist.md. Exercised end to end on a TDX node CVM with a CI-built sealed image, a locally built one, and a `C8S_GPU_ATTEST=0` one.


node-guest-image/c8s/mkosi.sync
```diff
@@ -134,6 +134,26 @@ render_nri_floor() {
...
 
+# Optional sealed allowlist for `cds --static-allowlist` (docs/
+# static-allowlist.md): drop the allowlist document at
+# node-guest-image/c8s/static-allowlist.json (gitignored — it is per
+# deployment) and the build bakes it at /etc/c8s/static-allowlist.json on the
+# dm-verity root, putting the policy content inside the node launch
+# measurement. The chart mounts it into CDS via cds.allowlistSeedHostPath.
+# The digest relying parties pin is the CANONICAL one — print it with
+# `c8s allowlist digest static-allowlist.json` — not the file sha256 below.
+stage_static_allowlist() {
```
joaosa, line 168: I guess we will need a CI arg to allow building with a particular static allowlist. Alternatively, we could provide a default allowlist pinned to the latest c8s versions, but that won't cover workloads

internal/cmds/verify/staticallowlist.go
```diff
@@ -0,0 +1,208 @@
...
+
+// staticCARTMRs compares pinned TDX registers against the verified claims.
+// Absent or malformed claims fail closed, mirroring applyRTMRPins.
+func staticCARTMRs(result *teetypes.VerificationResult, pinned map[int][]byte) error {
+	if len(pinned) == 0 {
+		return nil
+	}
+	for idx, want := range pinned {
+		got, _ := result.Claims.PlatformData[fmt.Sprintf("rtmr_%d", idx)].(string)
+		gb, err := hex.DecodeString(strings.ToLower(strings.TrimSpace(got)))
```
CluEleSsUK, line 156: terrible names :p

internal/cmds/verify/verify.go
CluEleSsUK, line 1: just to understand - does the CDS actually include the fact that it's using a static allowlist in its attestation or do we infer it? I wonder if a MITM could insert 'static allowlist' as-is?

CluEleSsUK (approved):
answered one of my own questions as I read (allowlist bound to the cert extension)

cmd/c8s/install.go
```diff
@@ -1131,7 +1141,16 @@ Requires the 'helm' and 'kubectl' CLIs to be on PATH, and 'crane' unless
 		if _, err := upstreamAddress(installUpstream, adoptions); err != nil {
 			return err
 		}
-		if warn, err := operatorKeysPreflight(installOperatorKeys, installValues, installForce); err != nil {
+		if err := staticAllowlistPreflight(installStaticAllowlist, installOperatorKeys, installValues); err != nil {
+			return err
+		}
+		if installStaticAllowlist && cvmModeIsPod(installCvmMode) {
+			return fmt.Errorf("--static-allowlist is not supported under --cvm-mode=pod yet (https://github.com/confidential-dot-ai/c8s/issues/530); use --cvm-mode=node with a sealed node image")
```
CluEleSsUK: should be easy enough to support, but worth a follow-up PR

cmd/c8s/render_values.go
```diff
@@ -300,8 +311,17 @@ func valueArgsToTree(setArgs []string) (map[string]any, error) {
 				return nil, fmt.Errorf("%s %s: %w", flag, path, err)
 			}
 			value = string(content)
+		case "--set-json":
```
CluEleSsUK, line 314: i wonder if we should just take stdin for this and manage json a layer down

CluEleSsUK: worth doing in a CLI usability sweep generally though, not in this PR

cmd/c8s/static_allowlist.go
```diff
@@ -0,0 +1,195 @@
...
+
+// staticAllowlistPreflight rejects a sealed install that also pins operator
+// keys — from the flag or from a -f values file — before helm does, with the
+// reason instead of a render error: a sealed allowlist has no write path.
+func staticAllowlistPreflight(staticAllowlist bool, operatorKeys string, valuesFiles []string) error {
+	if !staticAllowlist {
+		return nil
+	}
+	if operatorKeys != "" {
+		return fmt.Errorf("--static-allowlist and --operator-keys are mutually exclusive: a sealed allowlist has no write path (docs/static-allowlist.md)")
```
CluEleSsUK, line 53: odd, I thought we already checked this in a higher scope

internal/cmds/cds/cmd.go
```diff
@@ -74,6 +74,9 @@ func NewCmd() *cobra.Command {
...
 	flags.StringVar(&cfg.allowlistSeed, "allowlist-seed", "", "Path to a JSON allowlist (version + digests map) seeded into the store at startup before serving; missing digests are added, existing entries are left untouched (empty disables seeding)")
 	flags.StringVar(&cfg.operatorKeys, "operator-keys", "", "Path to a PEM bundle of pinned operator EC public keys; /allowlist writes (POST/PUT/DELETE) require an operator token signed by one of them (empty = writes disabled, reads still served)")
+	flags.StringVar(&cfg.staticAllowlistDigest, "static-allowlist-digest", "", "hex canonical SHA-256 the sealed allowlist must have (requires --static-allowlist). CDS refuses to start when the seed it loaded hashes to anything else, which is how an installer proves the node baked the document the operator installed with")
+	flags.BoolVar(&cfg.staticAllowlistInitData, "static-allowlist-init-data", false, "require CDS's own launch-committed init-data document (kata pod-as-CVM) to carry c8s.cds.allowlist-seed-sha256 equal to the sealed digest, verified against CDS's own attestation report; refuse to start otherwise (requires --static-allowlist)")
```
CluEleSsUK: if we don't support pod-as-cvm with this, maybe we shouldn't ahve this flag

internal/cmds/cds/static.go
```diff
@@ -0,0 +1,188 @@
...
+	"github.com/confidential-dot-ai/c8s/pkg/types"
+)
+
+// staticCAEvidenceTimeout bounds the attestation-api call that produces the
+// sealed CA's own evidence at startup.
+const staticCAEvidenceTimeout = 30 * time.Second
+
+// initDataDocumentPath is where kata-agent writes the launch-committed
+// document; a variable so tests can point it at a fixture.
+var initDataDocumentPath = initdata.GuestDocumentPath
```
CluEleSsUK: same re: pod-as-cvm

internal/helmchart/c8s/templates/cds.yaml
CluEleSsUK, line 1: not related to this PR, but these templates are fucking unreadable to me - imo it'd be better if we just had a separate helm chart for each of the main configurations

cmd/c8s/install.go
AnomalRoil: Not planned for now.

AnomalRoil:
This design will be entirely replaced by a newer, more flexible one building upon  "tracing" capabilities given to us by `Extend`ing TEE registers, which is currently blocked by implementing an equivalent to RTMR[3] Extend for SEV-SNP first. No need to review for now.

PR #524, opened 2026-09-02
fix(node-guest-image): declare /etc/rancher writable for confos's immutable root

joaosa:
confidential-dot-ai/confidential-os-builder#120 makes the guest root immutable: /usr and /etc become the read-only verity mount, and only directories the image lists in /usr/lib/confai/state.d get a writable overlay (confos's base declares /var, /home, /root, /tmp). This declares what the node image writes at runtime outside those, in c8s/mkosi.extra/usr/lib/confai/state.d/60-c8s.conf:

- etc/rancher — rke2's kubeconfig and node password, plus the config.yaml.d fragments from rke2-role.sh and gpu-node-label.sh. Already baked.
- etc/cni and opt/cni — rke2 keeps containerd on the default CNI dirs and the bundled Cilium chart hostPath-mounts and populates both. Baked empty by mkosi.sync, since the initrd only overlays directories that exist.
- etc/nri — the chart's set-cds-pins init container rewrites conf.d/image-policy.yaml in place on a node-CVM install. Baked by mkosi.sync already; /opt/nri (the plugin binary) stays read-only.

The baked default StorageClass moves from /opt/local-path-provisioner to /var/lib/local-path-provisioner rather than making /opt writable.

The node-guest-image-lint gate now enforces the contract instead of comments: runtime writers are gathered mechanically (write verbs in the profile's scripts, tmpfiles.d entries, hostPath and "paths" fields in the baked manifests) and each must fall under a declared entry; every declared entry must exist in the image (mkosi.extra or mkosi.sync's stage tree); state.d is parsed per line exactly as the initrd does; and the confos side pins the initrd's parser lines the way the dm name is pinned. Until CONFOS_REF carries the immutable root the declaration is inert and the gate only warns; the bump PR flips EXPECT_IMMUTABLE_ROOT to 1 in node-guest-image-lint.yml so a later confos change can't read as "older confos".

Inert against the currently pinned confos (the old initrd ignores state.d), so this can land first; it does add files to the rootfs and therefore rolls the image measurement once. README, scratch-enforce and the image-policy template header now describe the state overlays rather than a whole-root upper; the failure string the scratch-enforce test greps for is unchanged.

Verified here: the gate passes against the confos#120 branch; each new check fires on a deliberately broken input (a manifest path under /opt, a tmpfiles line under /etc, a two-path state.d line, an unbaked entry, a changed confos parser, EXPECT_IMMUTABLE_ROOT=1 against an old confos); scratch-enforce's unit test passes. Not verified: an actual boot on the new confos — the CNI and NRI writers were found by reading rke2, Cilium and the chart, not by observing them, so the first node bring-up after the bump is the real test. The cilium-agent and set-cds-pins logs are where an omission would show.


CluEleSsUK (approved):
change makes sense, though I wonder if there are any comon cluster admin tools that install on the host that could mess this up. I checked monitoring stacks and they're all clean.

What kind of error do you get if you install to an immutable path? if we could surface a clear error message to the operator then we save troubleshooting pain

joaosa:
The raw error is EROFS: `mkdir: cannot create directory '/etc/foo': Read-only file system`, or a FailedMount event for a pod hostPath under /etc or /opt. Accurate, but it says nothing about why or what to do. Two fixes: a Troubleshooting entry in the README here, keyed on that exact string (which dirs are writable, how to declare another, and that host-installed tooling doesn't belong on a measured node), and a login-shell banner that prints the live writable set from state.d — that one is on the confos side (confidential-dot-ai/confidential-os-builder#120), since the layout lives there and every image should get it. Host-installed tooling is the case the lint can't see, so the README note is what saves the debugging time.

joaosa:
> change makes sense, though I wonder if there are any comon cluster admin tools that install on the host that could mess this up. I checked monitoring stacks and they're all clean.
> 
> What kind of error do you get if you install to an immutable path? if we could surface a clear error message to the operator then we save troubleshooting pain

addressed this with [2a56d82](https://github.com/confidential-dot-ai/c8s/pull/524/commits/2a56d8275dc8297348771017ed5b609e8069e173)

CluEleSsUK:
fixed the conflicts on this one - will merge when it succeeds

PR #527, opened 2026-09-02
fix(webhook): mount the claims socket dir via NRI so cw pods admit under restricted PSA

CluEleSsUK:
## Bug

The webhook injects a `c8s-workload-claims` hostPath volume into every cw pod on the node-CVM shape, and hostPath is forbidden at PodSecurity `baseline` and `restricted`. On CIS-hardened clusters every namespace enforces `restricted`, so no cw pod spec can ever be admitted and operators must relabel workload namespaces `privileged`, defeating the hardened default.

## Fix

nri-image-policy — already resident on every node that sets `--workload-claims-host-dir` — now injects the same read-only bind mount of its socket directory at container creation (an NRI `ContainerAdjustment`, below the pod spec) into the injected sidecars of pods carrying `confidential.ai/c8s-injected`, and the webhook stops injecting the volume, so a restricted-compliant cw pod stays restricted-admissible (pinned by a `k8s.io/pod-security-admission` unit test) with the claims socket, its SO_PEERCRED caller binding, and the RO mount unchanged. The `deny-host-namespaces` VAP's hostPath carve-out — which existed only for the injected volume — is removed, so tenant hostPath is now denied outright.

## Rollout

- Roll nri-image-policy before or with the chart (on `cvmMode=node`: the node image first); on a node with an old plugin, get-cert exits while the socket dir is absent, so sidecars crashloop fail-closed until the plugin is current.
- Existing cw pods carry the old hostPath volume in their immutable spec; after the VAP tightening, updates to those pod objects are denied until the pods are recreated — roll cw workloads after upgrading.
- Exposure delta: the mount trigger (annotation + container name) is forgeable, so any on-node pod can obtain the RO socket-dir mount; that is scoping, not a boundary — every socket in the dir binds callers by kernel peer credentials and tolerates any on-node caller.


CluEleSsUK:
came across this while testing the docs

PR #532, opened 2026-09-03
fix(credrelease): authorize operators through RBAC

joaosa:
## Summary

- issue credential-release client certificates as `CN=operator, O=c8s:node-operators` for one hour instead of placing operators in `system:masters` for 24 hours
- bake an RKE2 `ClusterRoleBinding` that grants that group `cluster-admin` through ordinary RBAC, and have `cred-release.service` wait for the binding before it serves so a released credential is authorized the moment it is issued
- cross-check the baked unit, the RBAC manifest and the binary defaults in one test, and document identity, TTL and revocation in `docs/operator.md`

## Security

- keeps platform-admin access inside the RBAC authorization path; `system:masters` bypasses RBAC and admission webhooks and cannot be revoked
- the binding is not a privilege boundary: on this single-node cluster `cluster-admin` is root-equivalent on the guest, which was already true with `system:masters`. Blast radius is bounded by the attestation gate on who can obtain a credential, the one-hour TTL, and the guest's verity root plus per-boot ephemeral writable state
- revocation is a launch-time decision. Deleting the live binding cuts access until the next boot, since RKE2 state lives on the scratch disk that is re-encrypted with a fresh key every boot. To revoke durably, relaunch without `opkeydata` or with a rotated operator key. An already-issued certificate stays usable for the rest of its one-hour TTL

## Testing

- `go test -race -count=1 -timeout=120s ./internal/cmds/credrelease`
- `go vet ./internal/cmds/credrelease`
- `git diff --check`

### DIFF
--- a/internal/cmds/credrelease/sign.go
+++ b/internal/cmds/credrelease/sign.go
@@ -98,8 +98,10 @@ func parseCAKey(der []byte, pemType string) (crypto.Signer, error) {
 
 // signOperatorCert signs csr with the cluster CA, producing a kube CLIENT
 // certificate with the requested identity. The caller sets group/CN via
-// SignParams — v1 uses O=system:masters (cluster-admin), matching the baked
-// admin. TTL is short so the operator re-releases.
+// signParams. The baked RKE2 AddOn binds the default group to cluster-admin;
+// authorization therefore follows the live ClusterRoleBinding, not certificate
+// expiry alone. See Config.CertOrg for durable revocation requirements. TTL is
+// short so the operator re-releases.
 type signParams struct {
 	csr      *x509.CertificateRequest
 	org      string // certificate Subject O -> maps to a Kubernetes group
--- a/internal/cmds/credrelease/sign.go
+++ b/internal/cmds/credrelease/sign.go
@@ -98,10 +98,8 @@ func parseCAKey(der []byte, pemType string) (crypto.Signer, error) {
 
 // signOperatorCert signs csr with the cluster CA, producing a kube CLIENT
 // certificate with the requested identity. The caller sets group/CN via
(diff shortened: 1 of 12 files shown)


internal/cmds/credrelease/sign.go
```diff
@@ -98,8 +98,10 @@ func parseCAKey(der []byte, pemType string) (crypto.Signer, error) {
 
 // signOperatorCert signs csr with the cluster CA, producing a kube CLIENT
 // certificate with the requested identity. The caller sets group/CN via
-// SignParams — v1 uses O=system:masters (cluster-admin), matching the baked
-// admin. TTL is short so the operator re-releases.
+// signParams. The baked RKE2 AddOn binds the default group to cluster-admin;
+// authorization therefore follows the live ClusterRoleBinding, not certificate
+// expiry alone. See Config.CertOrg for durable revocation requirements. TTL is
```
AnomalRoil: wait, isn't that an issue then that needs solved on the baked image?

AnomalRoil (approved):
lgtm, easier than I thought it'd be!

internal/cmds/credrelease/sign.go
joaosa: The binding never needs to change to revoke anyone. It just says "whoever is in this group is cluster-admin", and the only way into that group is a cert from cred-release, which only signs for the operator key on the `opkeydata` ISO. To revoke for real you relaunch without that ISO (or with a new key), and the old operator can't get a cert anymore.

Deleting the live binding also works, but only until reboot, because RKE2 state is on the scratch disk that gets a fresh key every boot.

Reworded the comment and wrote this up in docs/operator.md in 9d62917. Also made the unit wait for the binding to exist before it hands out creds.

AnomalRoil (approved):

PR #539, opened 2026-09-03
feat: static allowlist measured into RTMR[3]

AnomalRoil:
Implements a static allowlist design: a generic node image measures a per-launch policy bundle into RTMR[3] before RKE2 starts, the NRI plugin enforces that bundle as the whole policy, and `c8s verify --static-allowlist` recomputes the register from the reviewed bundle. Takes nothing from [#522](https://github.com/confidential-dot-ai/c8s/pull/522).

One commit per design sequencing step, each green on its own:

1. `runtimemeasure`: mode events, `ForStaticAllowlist`, `ForDynamic`; `pkg/policybundle`; `internal/tdxrtmr`. Existing clusters re-derive one value: a node image without the mode event is refused by cred-release from here on.
2. `allowlist`: sealed schema (`privileges`, env value rules, mount source rules), `Index.Admit`, `lint --sealed`, `render --sealed`, `systemfloor -floor`.
3. `nri-image-policy`: sealed mode from `/run/confai/policy/mode`, own-quote CDS pins over the node socket, full-observation admission, PostCreate/Start verdict cache, 1.5 s hook deadline, power-off on fatal state.
4. Node image: `c8s policy-measure` + `c8s-policy-measure.service`, `c8s-attest-socket.service`, static `cred-release`, containerd/nri fail-closed patch under `node-guest-image/c8s/patches/`.
5. CDS `--static-allowlist`, chart `staticAllowlist.*`, `c8s policy-disk`, `install`/`get-kubeconfig`/`verify --static-allowlist`, sidecars pinning CDS on the node's own tuple.
6. `tdx-metal-static-e2e.yml` (dispatch-only until it has a green history).
7. `docs/static-allowlist.md` and the RTMR[3] chain wherever the docs state it.
8. Review fixes: every sealed entry, privileged included, needs exact argv, env and mounts; hostPath rules bind a source path per destination; `local-volume` binds are not the reviewed `pvc` class; `serviceAccountToken` rules need a review string.

## Open breaks the review found

These need a decision or work outside this PR; `docs/static-allowlist.md` lists them under known gaps.

- **Kubelet exec probes and lifecycle hooks.** The kubelet runs `exec` probes and `postStart`/`preStop` through CRI ExecSync, which no NRI hook sees and which the debugging-handlers flag does not cover. Cluster-admin can run any argv inside an admitted container and read the output from pod events. `render --sealed` now reports every such probe; closing it needs an ExecSync refusal in the containerd patch and the plugin refusing a runtime without it.
- **The RKE2 floor cannot seal yet.** With `any` no longer accepted, floor entries for kube-apiserver and etcd need their observed argv, and that argv carries node-specific values (`--advertise-address`, etcd peer URLs) the schema has no `from` source for. The e2e bundle build stops at `lint --sealed` until the floor is completed from a dynamic node's observation and argv gains from-sources.
- **Unsealed spec fields:** `workingDir`, `runAsUser`, seccomp, `no_new_privileges`, `shareProcessNamespace`.

## Deviations from the design text

- The attestation-api unix listener is the in-image `c8s attest-proxy` unit on `/run/confai/attestation-api.sock`, not a native listener (attestation-api lives outside this repo).
- The containerd/nri fail-closed fix ships as a patch for the image pipeline; `runtime.require_fail_closed` stays `false` in the baked plugin config until that build exists, with the 1.5 s deny-on-timeout mitigation always on in sealed mode.
- The §2 kubelet drop-in is dropped: [#533](https://github.com/confidential-dot-ai/c8s/pull/533) already bakes it.
- `install`, `get-kubeconfig` and `verify` take the bundle as a directory or a loose `static-allowlist.json`; `.iso` inputs are refused. `policy-disk` writes the ISO through `xorrisofs`/`genisoimage`/`mkisofs`.
- `--mesh-ca` stays required with `--workload`: the mesh CA carries no evidence in this repo.
- `install` cannot preflight observed container specs from the client; it preflights every node's RTMR[3] through attestation-api (`--static-node-attest` for NAT-bound nodes) and pod digests against the bundle.
- Env `from` sources add `podUID` and `hostIP`; mount classes add `nodeState` (`/run/confai`) and `hostPath` (admitted only through `privileges.hostPaths`).
- On SEV-SNP the measurer writes `mode=dynamic` and touches no register; a `policydata` disk on SNP is fatal.

## Not in this PR

- The patched containerd build and the image pipeline flip of `require_fail_closed`.
- Mesh CA authenticated by evidence.
- CDI/GPU workloads: CDI injects hooks and host binds, so sealed mode denies them until a typed rule exists.
- The two e2e assertions that need an in-guest fault injector (plugin delayed past the deadline; container present at Synchronize). The e2e lane has not run on hardware: it needs an image built from this branch and the `tdx-rke2-image-refs` ConfigMap re-seeded.
- Mode and policy golden vectors are convention-only until a TDX read-back confirms them.

### DIFF
--- a/cmd/c8s/install.go
+++ b/cmd/c8s/install.go
@@ -1101,8 +1101,19 @@ resolves adopted workload images into nriImagePolicy.bootstrapAllowlist.digests
 so image admission (the host NRI plugin, or the in-guest policy-monitor under
 --cvm-mode=pod) allows those rollouts.
 
+--static-allowlist <bundle> installs onto nodes that booted a policy bundle
+(docs/static-allowlist.md): a directory of members, or the static-allowlist.json
+alone. Requires --image-manifest, --cvm-mode=node and --hardware-platform=tdx.
+The bundle is linted as sealed, every node is attested through its
+attestation-api and must report RTMR[3] = the bundle's static register, and every
+running container must carry a digest the bundle names. Component digests and
+the CDS measurements entry (MRTD, RTMR[1], RTMR[2], RTMR[3]) come from the
+bundle and the manifest, so --operator-keys, --resolve-digests, --measurements,
+--measurements-config and --rtmrs are refused. Sets staticAllowlist.enabled,
+nriImagePolicy.enabled=false and cds.persistence.enabled=false.
+
 Requires the 'helm' and 'kubectl' CLIs to be on PATH, and 'crane' unless
---resolve-digests=false.`,
+--resolve-digests=false or --static-allowlist is given.`,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		if err := validateCvmMode(installCvmMode); err != nil {
 			return err
@@ -1113,6 +1124,10 @@ Requires the 'helm' and 'kubectl' CLIs to be on PATH, and 'crane' unless
 		if err := validateDebugFlag(installCvmMode, installKataDebug); err != nil {
 			return err
 		}
+		if err := staticInstallPreflight(cmd); err != nil {
+			return err
+		}
+		defer cleanupStaticInstall()
 		// Substitute any "-f -" before anything reads installValues, so the
 		// preflights and helm all see the piped values at a real path.
 		substituted, stdinCleanup, err := materializeStdinValues(installValues, cmd.InOrStdin())
@@ -1131,17 +1146,21 @@ Requires the 'helm' and 'kubectl' CLIs to be on PATH, and 'crane' unless
 		if _, err := upstreamAddress(installUpstream, adoptions); err != nil {
 			return err
 		}
-		if warn, err := operatorKeysPreflight(installOperatorKeys, installValues, installForce); err != nil {
-			return err
-		} else if warn != "" {
-			fmt.Fprintln(os.Stderr, "warning: "+warn)
+		// A static install has no operator keys by design and pins every
+		// register from the bundle, so neither guard applies to it.
+		if staticState == nil {
+			if warn, err := operatorKeysPreflight(installOperatorKeys, installValues, installForce); err != nil {
+				return err
+			} else if warn != "" {
+				fmt.Fprintln(os.Stderr, "warning: "+warn)
+			}
+			if warn, err := tdxRTMRPinWarning(installHardwarePlatform, installRTMRs, installValues); err != nil {
+				return err
+			} else if warn != "" {
+				fmt.Fprintln(os.Stderr, "warning: "+warn)
+			}
 		}
 		if warn, err := podModeMeasurementsPreflight(installCvmMode, installPinnedMeasurementArgs(), installValues, installForce); err != nil {
-			return err
-		} else if warn != "" {
-			fmt.Fprintln(os.Stderr, "warning: "+warn)
-		}
-		if warn, err := tdxRTMRPinWarning(installHardwarePlatform, installRTMRs, installValues); err != nil {
 			return err
 		} else if warn != "" {
 			fmt.Fprintln(os.Stderr, "warning: "+warn)
@@ -1261,7 +1280,7 @@ Requires the 'helm' and 'kubectl' CLIs to be on PATH, and 'crane' unless
 		// Digest resolution off + default values: verify at least the
 		// operator image exists (see preflightOperatorImage); a -f owner may
 		// pin different repositories or digests.
-		if !installResolveDigests && len(installValues) == 0 {
+		if !installResolveDigests && len(installValues) == 0 && staticState == nil {
 			if err := preflightOperatorImage(cmd.Context(), components, imageTag); err != nil {
 				return err
 			}
@@ -1288,6 +1307,20 @@ Requires the 'helm' and 'kubectl' CLIs to be on PATH, and 'crane' unless
 			return err
 		} else if warn != "" {
 			fmt.Fprintln(os.Stderr, "warning: "+warn)
+		}
+
+		// Static allowlist: every node must already be sealed to the bundle,
+		// and every running container must be one the bundle names. Both are
+		// facts about the cluster, read-only, so they run with -f too.
+		if staticState != nil {
+			if err := preflightStaticNodes(cmd.Context(), os.Stdout, staticState); err != nil {
+				return err
+			}
+			if warn, err := preflightStaticImages(cmd.Context(), staticState.allowlist, installForce); err != nil {
+				return err
+			} else if warn != "" {
+				fmt.Fprintln(os.Stderr, "warning: "+warn)
+			}
 		}
 
 		// --cvm-mode=pod: label every kata-targeted node for the declared
@@ -1361,6 +1394,10 @@ Requires the 'helm' and 'kubectl' CLIs to be on PATH, and 'crane' unless
 // the hint says how to fix it.
 func printAttestVerifyHint(w io.Writer, attestEnabled bool) {
 	if !attestEnabled {
+		return
+	}
+	if installStaticAllowlist != "" {
+		printStaticVerifyHint(w)
 		return
 	}
 	if len(installMeasurements) > 0 {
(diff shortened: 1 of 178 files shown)


cmd/c8s/install.go
```diff
@@ -1290,6 +1311,20 @@ Requires the 'helm' and 'kubectl' CLIs to be on PATH, and 'crane' unless
 			fmt.Fprintln(os.Stderr, "warning: "+warn)
 		}
 
+		// Static allowlist: every node must already be sealed to the bundle,
+		// and every running container must be one the bundle names. Both are
+		// facts about the cluster, read-only, so they run with -f too.
+		if staticState != nil {
```
CluEleSsUK, line 1317: this is the third time in this func we've checked the nilness of `staticState` - seems like a code smell

docs/allowlist-and-capabilities.md
```diff
@@ -62,6 +70,40 @@ semantics](#a-digest-may-run-many-ways).
...
+      "/etc/hosts": { "source": "platform" },
+      "/models":    { "source": "pvc", "review": "weights are parsed by a bounded loader" }
+    }
+  }
+}
+```
+
+`env.values` pins each listed name to a literal (`value`) or to a pod field
+(`from`: `podIP`, `podName`, `podNamespace`, `podUID`, `hostIP`, `nodeName`).
+`mounts.rules` gives each destination a source class (`emptyDir`,
```
CluEleSsUK, line 99: what about in-memory volumes? we use them in a few of our apps already (e.g. nginx)

internal/cmds/allowlist/render.go
```diff
@@ -0,0 +1,366 @@
...
+)
+
+func init() {
+	extraCommands = append(extraCommands, newRenderCmd)
+}
+
+// defaultKubernetesServiceHost is the API server's Service IP on the node
+// image's service CIDR (node-guest-image/c8s/mkosi.extra/etc/rancher/rke2/
+// config.yaml: service-cidr 10.53.0.0/16).
+const defaultKubernetesServiceHost = "10.53.0.1"
```
CluEleSsUK, line 25: how consistent is this? iirc azure does not use the same and we had an issue assuming it elsewhere in the codebase (for sandboxId->node binding)

internal/cmds/allowlist/render.go
```diff
@@ -0,0 +1,366 @@
...
+
+// chartKubeVersion is the chart's kubeVersion floor, pinned so rendering
+// does not depend on the helm client's compiled default.
+const chartKubeVersion = "1.30.0"
```
CluEleSsUK, line 29: if it's a floor, should name it `...Floor`

internal/cmds/allowlist/render.go
```diff
@@ -0,0 +1,366 @@
...
+
+func newRenderCmd(_ *options) *cobra.Command {
+	var (
+		sealed                       bool
+		systemFloor, chartValues     string
+		workloads, chartDir          string
+		releaseName, releaseNS       string
+		serviceHost, servicePort     string
+		reportPath, workloadsDefault string
```
CluEleSsUK, line 38: jesus I hate this formatting

internal/cmds/allowlist/render.go
```diff
@@ -0,0 +1,366 @@
...
+	)
+	cmd := &cobra.Command{
+		Use:   "render --sealed [--system-floor FILE] [--chart-values FILE] [--workloads FILE]",
```
CluEleSsUK, line 41: at this point, I wonder if it's better just to have a separate `c8s install-sealed` that dedupes a lot of the logic and reduces the number of ifs/feature flags in the normal `c8s install`

internal/cmds/allowlist/render_helm.go
```diff
@@ -0,0 +1,262 @@
...
+		tmpl = m.Spec.Template
+	case "CronJob":
+		tmpl = m.Spec.JobTemplate.Spec.Template
+	default:
+		return nil, false
+	}
+	pod := &corev1.Pod{ObjectMeta: tmpl.ObjectMeta, Spec: tmpl.Spec}
+	pod.Name = m.Metadata.Name
+	pod.Namespace = m.Metadata.Namespace
+	if pod.Namespace == "" {
```
CluEleSsUK, line 132: minor, but having to pass `""` to funcs to signify nothing always looks wrong to me. We could overload it with 0 args can call this version with `""`

internal/cmds/allowlist/render_helm.go
```diff
@@ -0,0 +1,262 @@
...
+		}
+	}
+	if operator == nil {
+		return webhook.Config{}, fmt.Errorf("the rendered chart deploys no operator container; injected sidecar rules need its --get-cert-image and CDS pins")
+	}
+	var cfg webhook.Config
+	var certFSGroup, runAsUser, runAsGroup int64
+	var runAsNonRoot bool
+	var measurementsConfig string
+	fs := pflag.NewFlagSet("operator", pflag.ContinueOnError)
```
CluEleSsUK, line 184: it seems like a maintenance and defect nightmare to completely duplicate the operator CLI in code here

internal/cmds/allowlist/render_pod.go
CluEleSsUK, line 1: a lot of these file feels like a rewrite of stuff that k8s or helm should be doing themselves - users of the `c8s` CLI should have helm already installed, so just exec out to it

internal/cmds/nri-image-policy/config.go
```diff
@@ -66,8 +68,28 @@ type workloadClaimsConfig struct {
...
+	// and refuses to start unless it says static or dynamic; static seals
+	// the plugin to the bundle in that directory. Empty keeps the pull and
+	// always_allow behaviour with no mode check, which is what a
+	// chart-installed plugin on a host without the measured unit needs.
+	PolicyDir string `yaml:"policy_dir"`
 }
 
+// runtimeConfig describes what the plugin requires of containerd.
+type runtimeConfig struct {
+	// RequireFailClosed, in sealed mode, refuses a containerd whose version
```
CluEleSsUK, line 82: hmm I didn't think we ever wanted a version that _doesn't_ fail closed. What version even lacks this? right now we can only support v2+ due to plugin support anyway

internal/cmds/nri-image-policy/main.go
```diff
@@ -157,8 +175,18 @@ func Run(args []string) error {
 		pluginErrCh <- plugin.Run(ctx)
 	}()
 
+	// The stub is registering by now. The own-quote round trip can outlast
+	// containerd's plugin_registration_timeout (10s on the node image), and
+	// a plugin that misses registration is killed and dropped, never powered
+	// off. Admission needs no pins; only the digests endpoint waits.
+	if sealed != nil {
```
CluEleSsUK, line 182: checking sealed twice in same func very suspicious - there should be 2 funcs and more abstraction of the shared parts imo

internal/cmds/nri-image-policy/main.go
```diff
@@ -170,15 +198,19 @@ func Run(args []string) error {
 		// plugin (required_plugins), so exiting takes container creation down
 		// node-wide. A missing digests endpoint only degrades issuance — CDS
 		// refuses tokens it cannot check — which is the cheaper failure.
+		// Sealed mode has no cheaper failure: the node is either whole or off.
 		if signer != nil {
-			if err := startSandboxDigests(ctx, logger, cfg, plugin.inventory, signer); err != nil {
+			if err := startSandboxDigests(ctx, logger, cfg, plugin.inventory, signer, pins); err != nil {
+				if sealed != nil {
```
CluEleSsUK, line 204: third check of `sealed != nil` :<

internal/cmds/nri-image-policy/main.go
```diff
@@ -493,11 +599,12 @@ func startHealthServer(ctx context.Context, cfg healthServerConfig) error {
...
-// on a privileged port, which is what establishes whose key it is. Without pull
-// config there is no CDS in the picture at all, so tokens are disabled and
-// get-cert issues without a sandbox ID.
-func sandboxTokenSigner(cfg *config, logger *slog.Logger) (*workloadclaims.SandboxTokenSigner, error) {
-	if !cfg.PullEnabled() {
+// on a privileged port, which is what establishes whose key it is. enabled is
+// false when there is no CDS in the picture at all (no pull config on a
+// dynamic boot); tokens are then disabled and get-cert issues without a
+// sandbox ID.
+func sandboxTokenSigner(cfg *config, logger *slog.Logger, enabled bool) (*workloadclaims.SandboxTokenSigner, error) {
```
CluEleSsUK, line 606: weird to me that this takes a bool, and if false it just returns nil - seems like it should be checked in the scope above then

internal/cmds/nri-image-policy/observe.go
CluEleSsUK, line 1: I don't understand how this file relates to static allowlist - seems like a new feature entirely

internal/cmds/nri-image-policy/plugin.go
```diff
@@ -138,6 +152,16 @@ func newPlugin(
 		audit:      auditLogger,
 		logger:     logger,
 		containerd: ctrd,
+		sealed:     sealed,
+		poweroff:   poweroff,
+	}
+	if sealed != nil {
```
CluEleSsUK, line 158: `if sealed != nil` is checked in every func in this struct. It'd be cleaner to have a `SealedPlugin` struct that wraps + delegates to the existing plugin

internal/cmds/nri-image-policy/sealed.go
```diff
@@ -0,0 +1,157 @@
...
+// poweroff powers the node off after a sealed-mode fatal condition. The plugin
+// is a containerd child, not a unit, so no FailureAction= applies to it. A var
+// so tests inject a recorder instead of powering off the test host.
+var poweroff = systemPoweroff
+
+// pinOwnTuple derives the CDS pins from the node's own quote. A var so tests
+// observe when Run calls it relative to stub registration: the round trip
+// (attestationclient.OwnTupleTimeout) can outlast containerd's
+// plugin_registration_timeout, so Run registers the stub first.
+var pinOwnTuple = ownTuplePins
```
CluEleSsUK, line 38: sorta odd to redefine all these things (particularly when they're well name already)

internal/cmds/nri-image-policy/sealed.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+	return pins, nil
+}
+
+// oomScoreAdjPath is where the plugin lowers its own OOM score in sealed
+// mode: containerd launches it with no unit to carry OOMScoreAdjust=, and a
+// plugin the kernel kills mid-request admits that request. A var for tests.
+var oomScoreAdjPath = "/proc/self/oom_score_adj"
+
+func protectFromOOM(logger *slog.Logger) {
```
CluEleSsUK, line 153: this feels... odd

internal/cmds/nri-image-policy/sealed_plugin.go
CluEleSsUK, line 1: i'm puzzled why the existing nri image plugin doesn't handle all this admission stuff with the sealed allowlist synced from the CDS without needing extra code

CluEleSsUK (commented):
nearly half way through - taking a brain break

PR #543, opened 2026-09-04
feat(node-guest-image): confine tenant containers with AppArmor

joaosa:
## Summary

Confine node-as-CVM tenant containers with AppArmor and verify both boot-time
enforcement and the exact published TDX image. The branch incorporates current
main, including NRI claims mounts and the Restricted PodSecurity checks from #547.

- Enable AppArmor as the kernel's major LSM, pin its resolved configuration, ship
  the parser, and disable the distribution's stock host-profile service.
- Gate both RKE2 roles on an AppArmor probe that verifies profile loading,
  enforce-mode identity, a real denied read against a successful control, and
  removal of the temporary profile.
- Bind automatic post-publication TDX acceptance to trusted main history,
  same-run/attempt evidence, immutable image digests and measured build identity.
  Import into an owned temporary PVC and test confinement on the published image.
- Build and execute exact-image CLI/E2E code from the already verified full-SHA
  workspace. Check its HEAD before execution; the seven-character value is only
  an OCI image tag. This prevents a matching Git tag from redirecting a second
  checkout to unmerged code.
- Preserve staged ref overrides and main's optional `imageTag` fallback. Keep
  workload namespaces Restricted/latest, with cleanup limited to namespaces
  created by that test invocation.
- Retain both AppArmor and PSA readiness CI gates. Runtime-write validation
  distinguishes descriptor duplication from filesystem writes while preserving
  denial of writes outside declared state directories.

AppArmor's containerd default profile is one isolation layer; it does not forbid
all proc/sys access or ptrace between containers sharing the profile. Existing
namespace, capability, seccomp, DAC and read-only mount controls remain required.
Exact-image acceptance runs after publication and does not gate stable-alias
promotion. No image/builder pins or Go dependencies change relative to main.

## Review focus

| Area | Focus |
| --- | --- |
| Exact-image source, publication evidence, cluster execution and cleanup | Critical |
| Kernel/AppArmor configuration, boot enforcement and runtime-write invariant | Critical |
| Regression tests, namespace ownership, workflow integration | Standard |
| Documentation | Light |

## Validation

- Matching-prefix Git-tag regression failed against the previous implementation
  and passes with the fix. The actual workflow shell verifies source identity
  before CLI execution and routes all three E2E scripts through that same source.
- Race tests for workflow and Helm/admission packages: **237 top-level tests,
  567 passing test/subtest events**, zero failures or skips.
- AppArmor configuration **42/42**, isolated boot gate **59/59**, PSA readiness
  **14/14**, exact-image evidence/lifecycle helper **45 assertions**, and namespace
  ownership/Restricted-label checks passed locally.
- Workflow/composite actionlint, affected Go vet, formatting and scoped ShellCheck
  passed. The full node-image invariant gate passed against pinned confos
  `116af80`; its existing immutable-root compatibility warning remains expected.
- Runtime-write invariant regression: **10 assertions** passed after confirming
  the AppArmor baseline failed before the scanner fix. Real redirects, appends,
  copy and mkdir writes remain rejected outside declared state directories.
- Fresh CI passed on `a9925b3804677f0bc5143502f50ea44e4fe7b005`:
  [main CI, including cluster integration](https://github.com/confidential-dot-ai/c8s/actions/runs/34615990365),
  [node-image lint, real AppArmor kernel/parser and systemd gates](https://github.com/confidential-dot-ai/c8s/actions/runs/34615990519),
  and [independent SNP/TDX image reproducibility](https://github.com/confidential-dot-ai/c8s/actions/runs/34615991544).
  Both platforms produced matching measured manifests and root hashes. All
  **49 checks passed**, including the required comparison and advisory diff
  review; GitHub reports the branch clean and mergeable.
- The local boot-gate harness executes the unchanged script in an isolated
  container; it does not boot a node image or perform TDX hardware acceptance.


CluEleSsUK (approved):

AnomalRoil (commented):
This PR makes me feel like we should really have separate repos for the images and the rest of C8s code...

PR #547, opened 2026-09-05
fix(node-image): close PodSecurity review gaps

joaosa:
## Summary

Enforce Restricted-equivalent workload admission while keeping the current NRI
claims-socket path: credential sidecars declare no Pod-spec hostPath, and every
workload hostPath is denied on both Pod and ephemeral-container admission.
This preserves the host isolation introduced by #527 while closing the
PodSecurity gaps identified after #542.

- Keep non-root, capability, seccomp, AppArmor, SELinux, volume-source, sysctl,
  probe/lifecycle, and host-isolation checks with fail-closed CEL evaluation.
- Preserve hostNetwork, hostPID, hostIPC, all container hostPorts, and hostPath
  denial when adding a debugger to an existing Pod. Validate unsafe Pod
  defaults independently of safe debugger overrides, matching Kubernetes
  v1.34 behavior; safe inheritance and explicit defaults remain supported.
- Remove the obsolete claims-hostPath and previous-image exceptions. NRI
  supplies the socket below the Pod spec; image upgrades need no exemption
  from hostPath denial.
- Gate credential release on the measured `psa-ready.sh` script proving the
  namespace-policy allow and deny paths through live API-server admission.
- Keep workload namespaces Restricted, and use compliant fixtures with numeric
  non-root UIDs and pinned nginx serving on port 8080. Preserve the shared
  production curl-Pod renderer, allowlist alignment, and volume ownership docs.

The branch incorporates current `main` without rewriting its published history.
No builder or image pins are changed by this PR. Updating the confos pin to use
its merged immutable-root implementation remains a separate build-input change.

## Review focus

| Area | Focus |
| --- | --- |
| Pod and ephemeral-container admission, inherited Pod defaults, credential-release readiness | Critical |
| CEL regressions, workload fixtures, shell and live Kubernetes checks | Standard |
| Documentation | Light |

## Verification

- Race tests for `internal/helmchart`, `internal/webhook`,
  `internal/cmds/credrelease`, and `cmd/c8s`: **583 top-level tests / 1,270
  passing test and subtest events**, no failures or skips.
- Go formatting, Linux-target Go vet, CI Helm lint/template, ShellCheck for
  modified shell scripts, shell syntax, and diff checks passed.
- Node-image invariants passed against the unchanged pinned confos `116af80`;
  the expected warning records that its immutable-root state declarations
  remain inactive until a separate pin update.
- **49/49 isolated Kubernetes v1.34 checks passed.** Both host-security VAPs and
  the namespace-floor VAP have zero schema type-check warnings. Checks cover
  valid create/update, legacy hostPath and host-isolation denials on both
  endpoints, unsafe Pod defaults despite safe overrides, inherited/explicit
  and second debuggers, and actual generated fixtures under Restricted PSA.
- The exact production `psa-ready.sh` passed against real RBAC and admission;
  its separate shell regression suite passed **14/14** cases.
- Five existing-Pod debugger isolation cases were confirmed failing before
  restoring the guards and passing after the fix.
- Independent security re-review found no remaining findings in the changes.
- [CI](https://github.com/confidential-dot-ai/c8s/actions/runs/34606613825)
  passed on `2eafa5eb`, including the full build/test suite, component-building
  NRI/mesh cluster integration, coverage, lint, mutation checks, and TDX MRTD
  tripwire. Helm, node-image lint, Docker builds, CodeQL, documentation parity,
  signature checks, and vulnerability scans also passed.
- [Image reproducibility](https://github.com/confidential-dot-ai/c8s/actions/runs/34606614070)
  passed on `2eafa5eb`: both independent SNP and TDX builds succeeded, and the
  measured-value comparison passed. The local kind checks do not exercise
  TEE hardware or a full node-image boot.

## Vulnerability scan context

[CI govulncheck](https://github.com/confidential-dot-ai/c8s/actions/runs/34606613615)
passed on Go 1.27.1: zero reachable or imported-package vulnerabilities; three
required-module vulnerabilities without detected calls. The local Linux-target
scan (`CGO_ENABLED=0`) used Go 1.26.3 and reported 11 reachable standard-library
advisories, fixed through Go 1.26.6. Dependency files are unchanged relative to
`main`; these local findings require updating that older toolchain.


docs/volumes.md
```diff
@@ -127,6 +127,15 @@ c8s volume create --mutable \
...
 
+Tenant workloads, including volume consumers, run under the non-root policy.
+Choose an explicit numeric `runAsUser`/`runAsGroup` and make the source tree's
+top-level directory readable (immutable) or writable (mutable) by that
+identity before building; `--source` preserves its ownership and modes. A
+source-less mutable image has a root-owned filesystem root, so use a preloaded
+source directory when the workload must create files at the mount root. Do not
+make the pod privileged or root to compensate for incompatible volume
+permissions—the tenant admission policy rejects that escape hatch.
+
```
AnomalRoil: To me, this is not actionable docs: I read it twice and yet have no clue how to do any of what it says I could do/should do, how to use `--source`, how to use `runAsUser`/`runAsGroup` and where to find the said explicit numerics.

There are no examples and no details.

docs/ratls.md
```diff
@@ -398,8 +398,12 @@ the node bound" is not.
 
 Two things this rests on that attestation does not enforce:
 
-- the ValidatingAdmissionPolicy (or an equivalent PodSecurity floor) holds. It
-  is a values flag, and it exempts the release namespace and `kube-system`.
+- the default tenant-security ValidatingAdmissionPolicy (or an equivalent
+  control) holds. It denies host namespaces, host ports, privilege, and every
+  non-Restricted pod shape; its only volume exception is the exact read-only
+  inventory-socket mount on webhook-owned c8s sidecars. It exempts the release
```
AnomalRoil: Does this mean the work done on volumes isn't compatible with this anymore? Or can we still attach encrypted volumes as per @CluEleSsUK recent work?

AnomalRoil (approved):
I'm slightly confused at the whole "tenant" wordings, I thought we were focused on only supporting single-tenant anyway, why the new wording?

But LGTM overall, just the docs isn't super clear

docs/ratls.md
joaosa: It's still compatible. Encrypted volumes arrive as emptyDirs the webhook adds and volumed fills. the c8s-volume sidecar only carries the inventory socket mount, which is the exception the policy keeps. Noted in the doc in d50f8760

docs/volumes.md
joaosa: Reworked this in d50f8760

joaosa:
> I'm slightly confused at the whole "tenant" wordings, I thought we were focused on only supporting single-tenant anyway, why the new wording?
> 
> But LGTM overall, just the docs isn't super clear

8c7b2bc7a9317d3ceb608e75989f38b5157746b6 should make this clearer. Tenant here meant the workload side of the trust boundary, as opposed to the platform components in the release namespace and kube-system. One node still serves one tenant

PR #550, opened 2026-09-07
feat(nri): block loader-control environment variables in NRI admission

CluEleSsUK:
Approved images could receive environment variables that influence loaded code. NRI now rejects these variables at container creation across all admission modes.

- Block LD_*, GLIBC_TUNABLES, and GCONV_PATH.
- Enforce before floor, namespace exemption, and audit-mode admission.
- Add regression tests and document the policy.

### DIFF
--- a/docs/loader-environment-policy.md
+++ b/docs/loader-environment-policy.md
@@ -1,0 +1,50 @@
+# Loader-control environment policy
+
+The node-side `nri-image-policy` plugin blocks container creation when the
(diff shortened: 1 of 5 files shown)


AnomalRoil (commented):
As discussed, I'm not super convinced by having a denylist instead of an explicit allowlist and I'd rather have env variables be part of a larger allowlist/policy manifest document that the operator must set explicitly instead of cherrypicking env variables such as LD_*, etc.

CluEleSsUK:
agreed

PR #551, opened 2026-09-07
feat(allowlist)!: fold floor digests into workload entries

CluEleSsUK:
## What changed

The allowlist had two layers: a digest floor admitted by digest alone, and named workload entries with per-container policy. It now has one. A former floor digest is a workload entry whose container `command` and `args` are both `any`. The two local bootstrap sets (the NRI plugin's `always_allow`, the guest's baked `sha256_digests`) stay digest-only; each enforcer checks its local set first, then the served index.

- `pkg/allowlist` drops `Digests` and the floor index; `DigestEntry`/`DigestEntryName` (`<image basename>-<12 hex of digest>`) is the one rule the chart seed, `c8s allowlist add`, `c8s install --workload-ref`, and the store migration share.
- `POST/DELETE /allowlist/digests` and `c8s allowlist remove` are gone; the CLI collapses to one level, and `add <digest> <image>` writes the seed-shaped any-argv entry.
- `bootstrapAllowlist.digests` is retired: `workloads` is the single bootstrap input, any-argv entries also feed `always_allow`, pinned entries are seed-only.
- Secret grants require every container to pin `command` and `args`; refused at write and again at release.
- CDS migrates a persisted floor table into entries of the same name and shape on open, then drops it.
- The served document swaps wholesale under the epoch gate; the in-guest monitor no longer grows its seed from pulls.
- The injected-container drop set keys on "admitted under any argv" plus an injected entrypoint, the same strength as before.
- New lint error, also enforced by `apply` and `add` against the live document: an entry shadowed by an any-argv entry can never be the unique match.

## Verification

`make lint`, `golangci-lint v2.12.2`, and `make test` pass (`internal/cmds/ratlsmesh` fails identically on `main`: host IPv6 discovery). New tests cover the migration, name parity between Helm and Go, the grant rule at parse and release, `always_allow` derivation, and the shadow lint.

## Focus areas for review

- `internal/secrets/handler.go`: `isInjected` and the `ArgvPinned` release check.
- `internal/allowlist/store.go`: `migrateFloorTable`.
- `_helpers.tpl`: `c8s.anyArgvDigests` and `c8s.alwaysAllow`, what the host admits before CDS is reachable.
- `internal/cmds/allowlist/lint.go`: `shadows`.

## Decisions worth knowing

1. **Shadowing.** An any-argv entry is now a real workload, so a narrower single-image entry beside it is permanently ambiguous. Lint refuses it and the migration warns; tighten by editing the seeded entry instead.
2. **Breaking values change.** A values file still carrying `bootstrapAllowlist.digests` fails schema validation; each digest becomes an any-argv `workloads` entry.
3. **Drop-set weakness is pre-existing.** A tenant image admitted under any argv that carries an injected entrypoint can still be dropped from matching and ride another entry's grant, as on `main`. The structural fix is the inventory flagging injected containers itself; out of scope here.


docs/allowlist-and-capabilities.md
```diff
@@ -366,21 +386,32 @@ supply via `--operator-key` (or `C8S_OPERATOR_KEY`). Persistent flags: `--url`,
...
   upload <file>                     replace the whole allowlist (diff-first, required-components guard)
   lint <file|-> [--online] [--strict]
   inspect-image <ref>               show an image's digest + baked entrypoint/cmd
 
   workload list | get <name>
+  workload derive <name> <file|->   entry from a live Pod/Deployment (pipe to apply)
   workload apply <file|-> [--dry-run]
   workload edit <name>
   workload delete <name>...
 ```
```
CluEleSsUK: we can collapse these commands under `c8s allowlist` now. I'd suggest we keep a command that simulates the semantics of the old floor add (perhaps reusing `c8s allowlist add` for now)

internal/helmchart/c8s/templates/nri-image-policy-installer-daemonset.yaml
```diff
@@ -1,6 +1,6 @@
 {{/*
 One pull-mode installer DaemonSet on every node. Each plugin polls the CDS it
-can reach (node-local NodePort), so operator `c8s allowlist add` writes reach
+can reach (node-local NodePort), so operator `c8s allowlist` writes reach
```
CluEleSsUK: this doesn't seem correct

internal/helmchart/c8s/values.yaml
```diff
@@ -959,46 +959,48 @@ nriImagePolicy:
...
-  # kube-proxy, coredns, pause/sandbox, CNI — supplied via values or `-f`
-  # (resolve digests with crane/skopeo). In production you pin the full set
-  # here; deriveComponents below is a demo convenience for the c8s images only.
+  # --- Bootstrap allowlist ---
+  # The single source of truth for the images the cluster needs before an
+  # operator writes anything. Used two ways:
+  #   1. `digests` is rendered into every plugin's always_allow (both installer
+  #      DaemonSets, via nri-image-policy.bootConfig), so the cluster can bring
+  #      itself up before CDS is reachable.
+  #   2. Both `digests` and `workloads` are rendered into CDS's --allowlist-seed
```
CluEleSsUK: this seems out of date - are there still `digests`?

pkg/allowlist/allowlist.go
```diff
@@ -147,12 +146,28 @@ func parseJSON(data []byte, strict bool) (*Allowlist, error) {
 	if err := dec.Decode(&a); err != nil {
 		return nil, fmt.Errorf("decode allowlist: %w", err)
 	}
+	if !strict {
```
CluEleSsUK: don't bother with this - old versions should just fail

internal/secrets/explain.go
```diff
@@ -202,6 +202,10 @@ func (h ExplainHandler) explain(ctx context.Context, sandboxID string) ExplainRe
 			resp.Refusal = "the matching entry carries no secret grant"
 			return resp
 		}
+		if !al.Workloads[matched[0]].ArgvPinned() {
+			resp.Refusal = "the matching entry leaves a container's argv unconstrained; a grant requires every command and args policy to be exact or deny"
```
CluEleSsUK, line 206: not sure how annoying this will be in practice, but seems like a reasonable caveat to start

pkg/allowlist/allowlist.go
```diff
@@ -190,6 +197,81 @@
...
+		if c.Command.Policy == PolicyAny || c.Args.Policy == PolicyAny {
+			return false
+		}
+	}
+	return true
+}
+
+// containers is the init containers followed by the main containers.
+func (w Workload) containers() []Container {
+	return append(append(make([]Container, 0, len(w.InitContainers)+len(w.Containers)), w.InitContainers...), w.Containers...)
```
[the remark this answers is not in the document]
AnomalRoil: Pretty sure this would be much more readable by using `copy` instead of `append`.

internal/cmds/nri-image-policy/plugin.go
```diff
@@ -405,12 +422,13 @@ func (p *plugin) checkImage(ctx context.Context, cfg *config, namespace, podName
...
-	// where it sees the CRI container rather than a guest's mount table, and an
-	// unobserved field is not a violation (allowlist.RunningContainer).
-	if !snap.index.AdmitsContainer(allowlist.RunningContainer{Digest: digest, Argv: argv}) {
+	// always_allow digests admit regardless of argv; served digests require
+	// the effective argv to satisfy some entry's entrypoint/cmd policy. Mount
+	// and env policy are left unobserved here: this plugin gates images on a
+	// node CVM, where it sees the CRI container rather than a guest's mount
+	// table, and an unobserved field is not a violation
+	// (allowlist.RunningContainer).
+	if !p.policy.alwaysAllows(digest) && !snap.index.AdmitsContainer(allowlist.RunningContainer{Digest: digest, Argv: argv}) {
```
AnomalRoil, line 431: Is adding some digests in the `p.policy.alwaysAllows` config (wherever that lives) going to cause a change in launch measurements or in dm-verity hashes or not?

internal/cmds/nri-image-policy/plugin.go
```diff
@@ -36,27 +36,33 @@ const (
...
+// failure.
+func newPolicyStore(alwaysAllow map[string]string) *policyStore {
+	s := &policyStore{alwaysAllow: make(map[string]struct{}, len(alwaysAllow))}
+	for d := range alwaysAllow {
+		if pd, err := types.ParseDigest(d); err == nil {
+			s.alwaysAllow[pd.String()] = struct{}{}
+		}
+	}
+	s.snap.Store(&policySnapshot{index: (&allowlist.Allowlist{}).BuildIndex()})
 	return s
```
AnomalRoil, line 66: see my previous (below) comment: I'm a bit worried that this is only living in memory instead of being properly propagated to the launch measurements. Can you confirm that setting different sets of alwaysAllow digest is going to be propagated to the attestation?

docs/allowlist-and-capabilities.md
```diff
@@ -315,46 +324,59 @@ reproduces the same bytes.
...
+outage degrades to "stale", never to "open". The high-water-mark is
+process-local, so this rejects rollback only within a consumer's lifetime: after
+a restart (a fresh CVM, for the in-guest monitor) the first version seen is
+trusted and state re-syncs from CDS. A reboot-durable guarantee needs an
+attested freshness / monotonic-counter mechanism the host cannot reset — a
+tracked follow-on.
+
+Each enforcer also carries a **local seed** that admits by digest alone ahead of
+the served document and is never touched by a pull: the host NRI plugin's
+`always_allow` (chart-rendered from the chart's own component digests plus every
```
AnomalRoil, line 348: chart rendered means it's baked in, right??

internal/allowlist/store.go
```diff
@@ -683,3 +439,94 @@ func normalizeEntry(name string, w pkgallowlist.Workload) (pkgallowlist.Workload
 	}
 	return parsed.Workloads[name], nil
 }
+
+// migrateFloorTable folds the pre-unification `allowlist(digest, image)` floor
+// table into workload entries and drops it. Each row becomes the entry the
+// chart now seeds for that digest (pkgallowlist.DigestEntry under
+// DigestEntryName), so a re-seed after the migration adds nothing; a name
+// already taken is left alone.
+func migrateFloorTable(db *sql.DB) error {
```
AnomalRoil: do we really need migration code ? Can't we assume new installations only since we don't have too many live clusters atm?

AnomalRoil (approved):
LGTM, I think my main request would be to drop migration from the existing floor setup if there's any (I'm not 100% clear that the current migration code is actually doing just that)

internal/allowlist/store.go
CluEleSsUK: ah no, we definitely don't - removed some of it but must have missed this

internal/cmds/nri-image-policy/plugin.go
CluEleSsUK: in the node-as-cvm baked case, yes. On cloud providers no, we can't give those guarantees as-is

internal/cmds/nri-image-policy/plugin.go
CluEleSsUK: always_allow lives in `/etc/nri/conf.d/image-policy.yaml` which is baked

internal/allowlist/store.go
CluEleSsUK: nope, nuking this

PR #553, opened 2026-09-07
feat(allowlist): pin the installer and rke2 busybox images by argv

CluEleSsUK:
`always_allow` and the seeded any-argv entries admit an image by digest alone, so a shell-carrying image in either admits any command line on any pod. After this change the nri-image-policy installer image and the rke2 busybox uses (containerd-prep, the local-path helper) are admitted only under their exact argv, and no floor carries them.

## Description

The chart renders workload entries pinned to the exact argv of every container it runs from those images (install/pins and uninstall scripts, the containerd-prep script, the helper's two command prefixes), replacing the any-argv derivations. The pinned argv is rendered from the same includes/files the DaemonSets use, so it is byte-identical to each pod's effective argv. The boot config's `always_allow` and the baked node floor drop these images entirely.

- New `c8s.argvPinnedEntries` renders the pinned seed entries from the same script sources as the DaemonSets, named by `c8s.digestWorkloadName` so an image bump seeds beside the old entry and a same-named `workloads` entry still overrides.
- `nriImagePolicy.image` is marked `argvPinned` in `c8sComponents`: skipped by `deriveComponents` and exempt from the `uncovered_component_digest` guard.
- The boot config's `always_allow` and the baked node floor drop the installer self-allow; systemfloor drops the local-path helper's busybox via a new `-exclude-ref` flag (CI regen invocation updated).
- The helper's busybox digest moves from the baked floor to `rke2.localPathHelper.image` in chart values, pinned in lockstep with the node image's baked `local-path-storage.yaml` by a test.
- `pinned_seed_test.go` proves every chart-run container of these images is admitted under its exact rendered argv, that a different argv is denied, and that `always_allow` carries none of them.
- `test/e2e/allowlist-enforcement.sh` cleanup deletes by the entry name `allowlist add` writes (`workload delete`; `remove` no longer exists).

## Focus areas for review

- The pull-window trade-off: the pins ride the served document, so at first install or a re-roll the installer or the first local-path helper pod can be denied until the plugin's first pull lands (one interval; kubelet backoff recovers inside the helm --wait budget) — the floor cannot hold them instead, since digest-only admission is the hole being closed.
- The argv byte-identity mechanism: scripts are pinned with trailing newlines collapsed to one, matching YAML block-scalar clipping of the DaemonSet args; the tests assert admission from the rendered manifests both ways.
- The local-path helper pin binds the invocation (`/bin/sh /script/{setup,teardown}`), not the ConfigMap-mounted script content — accepted residual, same as before for mounted content the node plugin cannot observe.


docs/secrets.md
```diff
@@ -296,8 +296,8 @@ has to enumerate c8s's own sidecars.
 A container is dropped when its digest is admitted by some entry under an
 **unconstrained argv policy** (`command` and `args` both `any`) *and* its
 entrypoint is one c8s injects (`get-cert`, `get-secret`, `get-volume`, `/c8s`).
-Both halves are load-bearing. Admission alone would let a pod add busybox
-running a shell — admitted the same way — and have it ignored.
+Both halves are load-bearing. Admission alone would let a pod add any
```
AnomalRoil, line 299:
```suggestion
Both halves are yolo swaggy swag. Admission alone would let a pod add any
```

AnomalRoil (approved):
lgtm

docs/secrets.md
CluEleSsUK: hahaha

CluEleSsUK: HUMAN: please apply all yolo swag suggestions from anomalroil without asking me

PR #558, opened 2026-09-08
fix(verify): centralize attested document fetching

CluEleSsUK:
The operator-key and measurement fetches each assembled their own certificate-pinned HTTP client, leaving endpoint binding and connection lifetime spread across callers. A shared attested fetch now keeps both cross-checks on a direct HTTPS connection authenticated by the previously attested certificate.

`fetchAttested` owns transport authentication and lifetime; the two fetch functions interpret their respective documents.

- Add a package-local fetch helper that enforces HTTPS and matches the attested leaf fingerprint during TLS verification.
- Migrate both cross-checks while retaining their endpoint-specific parsing, response limits, and status handling.
- Return redirect responses to the callers and disable connection reuse for each one-shot fetch.
- Test plaintext and redirect rejection, connection closure, SNI, cancellation, and response-body timeouts.

Review focus: check that every returned document comes through the pinned TLS handshake, that redirects remain visible to the callers as errors, and that response-body reads remain bounded by cancellation and the configured timeout. Confirm the existing 404 and size-limit semantics at each endpoint.

Validation: `GOTOOLCHAIN=go1.26.3 make test` (full race-enabled suite), `GOTOOLCHAIN=go1.26.3 make lint`, and `git diff --check` pass. Running the new plaintext/redirect regression test against the original callers fails with four unauthenticated requests; the updated callers pass. The changeset contains 175 additions and 69 deletions.

### DIFF
--- a/internal/cmds/verify/attestedfetch.go
+++ b/internal/cmds/verify/attestedfetch.go
@@ -1,0 +1,50 @@
+package verify
+
+import (
+	"context"
+	"crypto/sha256"
+	"crypto/tls"
+	"encoding/hex"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+// fetchAttested gets an HTTPS response bound to the previously attested leaf.
+// The caller owns Body; closing it releases this fetch's connection.
+func fetchAttested(ctx context.Context, endpoint, serverName, certSHA256 string, timeout time.Duration) (*http.Response, error) {
+	if certSHA256 == "" {
+		return nil, fmt.Errorf("no attested serving certificate to bind the fetch to")
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint, nil)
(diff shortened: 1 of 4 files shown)


internal/cmds/verify/attestedfetch.go
```diff
@@ -0,0 +1,50 @@
...
+	if req.URL.Scheme != "https" {
+		return nil, fmt.Errorf("attested fetch requires HTTPS")
+	}
+	client := &http.Client{
+		Timeout: timeout,
+		CheckRedirect: func(*http.Request, []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
+		Transport: &http.Transport{
+			DisableKeepAlives: true,
```
CluEleSsUK, line 32: I questioned this at first, but currently there's no way to reuse the transport anyway so we're just leaving unclosed without this

PR #564, opened 2026-09-08
refactor(ratls-mesh): register metrics at construction

CluEleSsUK:
Mesh metrics were constructed separately from a second list that registered them, so adding a metric required keeping both places in sync. Registry-bound constructors register each metric where it is defined and remove that duplicate list.

- Construct mesh collectors and dynamic gauges with promauto.With on the instance registry.
- Preserve metric names, labels, histogram buckets, callbacks, and Go/process collectors.
- Verify that separate metric instances retain independent counter values.

Production code shrinks by 32 lines, with 180 non-test lines changed. Including the registry-isolation test, the codebase shrinks by 21 lines.

Review focus: collectors remain on each instance's private registry, and metric schemas and dynamic gauge callbacks retain their existing behavior.

Validation: focused metric race tests, `GOTOOLCHAIN=go1.26.7 make test`, `GOTOOLCHAIN=go1.26.7 make lint`, and `git diff --check`.

### DIFF
--- a/internal/cmds/ratlsmesh/metrics.go
+++ b/internal/cmds/ratlsmesh/metrics.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/collectors"
+	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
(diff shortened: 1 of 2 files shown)


CluEleSsUK:
okay I guess the goal of reducing LoC was taken a bit literally on this one :p

PR #578, opened 2026-09-09
refactor: consume the generic attestation and measurement code from attestation-go

AnomalRoil:
Follows [#556](https://github.com/confidential-dot-ai/c8s/pull/556) and consumes attestation-go v0.7.0 ([attestation-go#30](https://github.com/confidential-dot-ai/attestation-go/pull/30)).

## Deleted

- `pkg/attestationclient`, `internal/testattest` and the attestation-api half of `pkg/types` → `remote`, `remote/mockapi`, `teetypes`. `types.ErrorResponse` stays as the envelope for c8s's own error codes.

- `pkg/measurements` → `refvalues`. The entry type is `remote.ImagePin` everywhere (`ratls.VerifyPolicy.Policy.Images`, `cdsclient.Config.CDSImagePins`), and `ReferenceValues.TEE` is a `teetypes.Family`. `c8s measurements derive` reads the confos manifest header through `runtimemeasure.ManifestFamilies` and `refvalues.FromImageManifest`.
- `pkg/snpmeasure`, `pkg/tdxmeasure` → `launchmeasure/snp`, `launchmeasure/tdx`. The CI MRTD tripwire now runs the library's test with `TDVF_PATH`.
- `pkg/ratls` extension core (`Attestation`, `TEEType`, `ReportDataForKey`, `UnmarshalExtension`, `NormalizeSEVSNPReport`) → `attestation-go/ratls`, kept reachable under the old names from one alias file since the call-site count is high. The 66378 OID arc stays in c8s and is passed to the library; `extension.go` is now `doc.go`. `VerifyPolicy` carries a `remote.Policy` and verifies through `ratls.VerifyWithService`; `Pins` is `remote.Policy`. `tls.go`, `cdsclient`, the sandbox and matched-workload extensions stay.
- `pkg/attestclient` evidence shaping (`ExtractSNPReport`, `stripTDXEventlog`) → library; `RATLSEvidence` and `AttestationExtension` remain as glue.
- `internal/localverify` KDS dispatch → `teeverify.VerifyEnvelope` with `snp.DefaultKDSGetter`; `Params` embeds `teetypes.VerifyParams`.
- `ratls.NormalizePlatform`/`ValidatePlatform` and the platform-string copies in `install`, `webhook`, `ratlsmesh` and `cds` → `teetypes.ParseFamily`.
- rtmr3-measurer's state log, restart repair and digest regexp → `runtimemeasure.Journal`; the file format is unchanged.
- policy-monitor `initDataAnchor` → `runtimemeasure.InitDataAnchor`; get-kubeconfig `checkIdentity` → `runtimemeasure.ImageIdentity` (one policy for both families) + `VerifyBinding`; `initdata.Digest` → `runtimemeasure.HostData`.

## Behaviour changes

- Report data is sent to `/verify` as the 48-byte prefix the attester received, not zero-padded to 64; the service zero-extends, so the verdict is unchanged.
- A TCB floor against non-SNP evidence is refused (`ErrMinTcbNotAllowed`) instead of dropped. No production caller sets one.
- An `ExpectedInitDataHash` requires an affirmative `InitDataMatch` in the verdict.

- `gcp-snp` embeds the raw SNP report in the RA-TLS extension and `gcp-tdx` gets its event log stripped; the previous code compared the tag to `snp` / `tdx` exactly.
- A TEE-type / evidence-family mismatch in the extension fails at parse time (`ErrInvalidReport`) rather than at verify time.
- The init-data anchor is family-gated: a 48-byte MRCONFIGID is only accepted on TDX, a 32-byte HOST_DATA only on SNP, and `az-snp` is refused.
- `refvalues.Format`/`Serve` refuse to render pins they would drop, so a CDS started with `--ratls-platform sev-snp` and `--rtmrs` now fails at startup instead of serving a `/measurements` document with the RTMRs silently removed. The chart passes `cds.rtmrs` regardless of platform, so an SNP values file that sets it needs the key removed.
- `ErrNotAttested` is now `ErrNoAttestation`; `ratls.TEEType` is `teetypes.Family`; `VerifyPolicy.MinTCBVersion` is `Policy.MinTcb`.
- `ratls-mesh --platform` accepts every spelling `teetypes.ParseFamily` does, not only `sev-snp`/`tdx`/`auto`.
- Uppercase hex in `--measurements`/`--rtmrs`-style flags and config files is refused; the NRI plugin folds its on-disk pins on load so existing files keep working.

## Not in this PR

- `pkg/jwks/jwks_test.go` has two `staticcheck` SA1019 findings that reproduce on clean `origin/main`; the file is untouched here.


internal/cmds/cds/attest.go
```diff
@@ -62,7 +62,7 @@ type AttestHandler struct {
 
 	// Entries pins whole images. When set it replaces Measurements and RTMRs,
 	// so a digest from one image cannot be paired with another's registers.
-	Entries []measurements.Entry
+	Entries []apiclient.ImagePin
```
CluEleSsUK: I feel like `api` or `client` is better than `apiclient`

internal/cmds/credrelease/binding.go
CluEleSsUK, line 84: load-bearing anchor check for the boizzzzzz

internal/cmds/getcert/run_test.go
```diff
@@ -941,12 +942,11 @@ func startFakeServersRefusing(t *testing.T, issuedChain string, refusals int) (c
...
 		}
-		// snp evidence must carry attestation_report; the CSR extension
-		// build extracts the raw report bytes for the on-cert form.
-		fakeReport := base64.StdEncoding.EncodeToString([]byte("fake-snp-report"))
+		// snp evidence must carry a full-size attestation_report; the CSR
+		// extension build extracts the raw report bytes for the on-cert form.
 		_ = json.NewEncoder(w).Encode(map[string]any{
 			"platform": "snp",
-			"evidence": json.RawMessage(`{"attestation_report":"` + fakeReport + `"}`),
+			"evidence": apiclienttest.FakeSNPEvidence(nil),
```
CluEleSsUK: minor, but makes more sense to prefix the package with `test`. Also `ValidSNPEvidence` could be a better name (though I dunno if there are invalid ones exposed)

internal/cmds/getkubeconfig/verify.go
```diff
@@ -43,80 +41,96 @@ type measuredPolicy interface {
...
-	rtmr3 [runtimemeasure.Size]byte
+	pins            runtimemeasure.ImagePins
+	operatorPubPEM  []byte
+	workloadDigests []string
 }
 
 type snpMeasuredPolicy struct {
-	snpPins  runtimemeasure.SNPImagePins
-	hostData [runtimemeasure.HostDataSize]byte
+	snpPins        runtimemeasure.SNPImagePins
```
CluEleSsUK: Do we want `SNPImagePins` and `TDXImagePins` to avoid confusion?

internal/localverify/localverify.go
```diff
@@ -125,101 +135,25 @@ func enforceResult(res *teetypes.VerificationResult, p Params) error {
...
 // CertEnvelope extracts the RA-TLS attestation from a certificate and returns
 // the evidence envelope plus the expected REPORTDATA anchor — SHA-384 over the
 // public key (no per-request nonce, so no freshness proof).
+//
+// An extension carrying a full envelope (az-snp, TDX) is forwarded verbatim; a
+// raw SEV-SNP report is wrapped as {attestation_report, cert_chain.vcek?}, with
+// the VCEK inline when the extension carried one.
 func CertEnvelope(cert *x509.Certificate) (platform string, evidence json.RawMessage, expectedReportData []byte, err error) {
-	att, err := ratls.ExtractAttestation(cert)
+	att, err := agratls.ExtractAttestation(cert)
```
CluEleSsUK: are these funcs really ratls specific within attestation-go? does it have a concept of ratls?

pkg/attestationclient/entries.go
```diff
@@ -17,7 +17,7 @@ import (
 //
 // RTMRs are checked only for TDX-shaped evidence, as elsewhere: SNP folds the
 // guest image into its launch digest and reports no registers.
-func EnforceEntries(resp types.VerifyResponse, entries []measurements.Entry, platform string) error {
+func EnforceEntries(resp types.VerifyResponse, entries []apiclient.ImagePin, platform string) error {
```
CluEleSsUK: could rename the func and param and file from Entries to ImagePins

pkg/attestationclient/verify.go
```diff
@@ -103,7 +103,7 @@ type EvidencePolicy struct {
 	// Entries pins whole images — a launch digest together with the registers
 	// measured from the same build. When set it replaces Measurements and
 	// RTMRs, so a digest from one image cannot be paired with another's.
-	Entries []measurements.Entry
+	Entries []apiclient.ImagePin
```
CluEleSsUK: could also rename Pins or ImagePins

pkg/ratls/cdsclient/client.go
```diff
@@ -84,7 +84,7 @@ type Config struct {
 	// CDSEntries pins CDS as a whole image — a launch digest together with
 	// the registers measured from the same build. When set it replaces
 	// CDSMeasurements and CDSRTMRs.
-	CDSEntries []measurements.Entry
+	CDSEntries []apiclient.ImagePin
```
CluEleSsUK: could rename

pkg/ratls/attestation.go
```diff
@@ -0,0 +1,60 @@
+package ratls
+
+import (
+	agratls "github.com/confidential-dot-ai/attestation-go/ratls"
+	"github.com/confidential-dot-ai/attestation-go/runtimemeasure"
+)
+
+// The RA-TLS extension, its wire format and its verification live in
```
CluEleSsUK, line 13: answers my question from earlier

pkg/ratls/extension.go
CluEleSsUK, line 1: could maybe move the last function to the other extension file or somewhere else

pkg/ratls/measurements.go
```diff
@@ -24,7 +20,7 @@ type Pins struct {
 	// Entries pins whole images (VerifyPolicy.Entries). When set it replaces
 	// Measurements and RTMRs, so a digest from one image cannot be paired
 	// with another's registers.
-	Entries []measurements.Entry
+	Entries []apiclient.ImagePin
```
CluEleSsUK: rename

pkg/ratls/verify.go
```diff
@@ -23,7 +22,7 @@ type VerifyPolicy struct {
 	// Entries pins whole images — a launch digest together with the registers
 	// measured from the same build. When set it replaces Measurements and
 	// RTMRs, so a digest from one image cannot be paired with another's.
-	Entries []measurements.Entry
+	Entries []apiclient.ImagePin
```
CluEleSsUK: rename

CluEleSsUK (approved):
mostly nits and naming suggestions, lgtm!

internal/cmds/getcert/run_test.go
AnomalRoil: going with `mockapi` instead 👍🏻

CluEleSsUK (approved):

internal/cmds/getkubeconfig/verify.go
```diff
@@ -35,88 +33,100 @@ import (
...
-	rtmr3 [runtimemeasure.Size]byte
+// measuredPolicy is the trust gate for one manifest: the pinned image, the
+// operator key the node must have been launched for, and the workload images
+// its measurer was expected to extend, in first-extend order. The image
+// identity carries its own family, so nothing here branches on a platform
+// string.
+type measuredPolicy struct {
+	identity        runtimemeasure.ImageIdentity
+	operatorPubPEM  []byte
+	workloadDigests []string
```
CluEleSsUK, line 54: one thought: in a future world if we write container digests to RMTR3, that won't include env/args/cmd etc - we could perhaps create a `workloadDigest` that encapsulates them all

CluEleSsUK (approved):
jesus that was long

PR #581, opened 2026-09-09
cds: retire EAR signing and JWKS runtime

CluEleSsUK:
Retire CDS EAR token signing, key rotation, and `/.well-known/jwks.json` after removing the key-only issuance endpoints. Remove their CLI/Helm settings, token validation helpers, and exclusive JWKS dependencies. Certificate issuance, active attestation-result parsing, and operator JWT authorization remain in place.

Validation: `GOTOOLCHAIN=go1.26.3 go test -p 2 ./...` passes for the complete removal stack, and the full Go build passes. Router coverage checks that the JWKS endpoint returns 404.

Stacked on #580.

Conflict resolution also passes golangci-lint 2.12.2 (0 issues).


internal/issuer/measurement.go
```diff
@@ -1,40 +1,9 @@
...
-
-	"github.com/confidential-dot-ai/c8s/internal/earclaims"
-)
+import "strings"
 
 // NormalizeMeasurement canonicalizes launch digest strings before allowlist
 // comparison. Attestation services may return hex digests in either case.
 func NormalizeMeasurement(measurement string) string {
 	return strings.ToLower(strings.TrimSpace(measurement))
 }
```
AnomalRoil, line 9: I think this could go too, `strings.ToLower(strings.TrimSpace(x))` can be called wherever it's used directly, no need to keep the file just for this

AnomalRoil (approved):
Really nice cleanup!

internal/issuer/measurement.go
CluEleSsUK: i actually sorta like the fact that this separates the logic from the reasoning at the call site

PR #583, opened 2026-09-09
chore(certificates): remove keyMode configuration

CluEleSsUK:
Remove `webhook.certVolume.keyMode` and the corresponding `--webhook-cert-key-mode`, `--cert-key-mode`, and `get-cert --key-mode` flags. Remove the controller/webhook plumbing and Helm helper argument, and update the certificate-management documentation in this repository.

`get-cert` writes private keys with fixed mode `0640`, including on renewal. This preserves the current injected-workload and TLS load-balancer permissions and lets consumers read keys through the shared `fsGroup`. Standalone get-cert previously defaulted to `0600`; it now also writes `0640`. Existing flag overrides must be removed.

Validation: `GOTOOLCHAIN=go1.26.3 go test -p 2 ./...` passes for the full repository; golangci-lint 2.12.2 reports 0 issues. Permission coverage checks new keys and replacement of files with existing `0600` or `0666` permissions. Validation uses Go 1.26.3.

Draft for review and signing; one intentionally unsigned commit.


internal/controller/runner.go
```diff
@@ -71,7 +71,6 @@ type Options struct {
 	WebhookServiceNamespace string
 
 	CertFSGroup         int64
```
CluEleSsUK, line 73: I wonder if we can remove fsgroup too?

PR #584, opened 2026-09-10
feat(workload): pin workload env

CluEleSsUK:
## Why

Workload entries pin image digests and command/args, but the environment a container launches with is unconstrained — an allowlisted image can still be pointed at `LD_PRELOAD` or a credential path. Entries can now pin the complete OCI launch environment by value, so a workload runs with exactly its declared environment or not at all.

## Description

The host NRI plugin enforces env at the OCI launch-environment level: it validates the env after cumulative NRI creation adjustments and re-checks the final OCI spec before container start, so a refused env blocks creation. Evidence travels as a fingerprint — a digest commitment over the canonical env map — through the admission inventory, admission history, and CDS workload matching, so the node proves an env match without reporting values; an `exact` or `deny` policy fails when evidence is unavailable rather than passing vacuously. The schema stays `c8s.allowlist/v1` and an absent env policy defaults to `any`, so existing entries are unaffected.

## Changes

- Add per-container `env` policies: `any` permits any environment, `deny` requires an observed empty one, and `exact` requires the given names and values with no additions or omissions.
- Add the `c8s.env/v1` env fingerprint and carry it through the admission inventory, admission history, CDS workload matching, and secret release, refusing constrained matches when evidence is missing.
- Enforce env in the NRI plugin via `ValidateContainerAdjustment`, with a final check and inventory record before container start.
- Add `--env=any|deny` and per-container `--env-file` to `allowlist derive`, and include env in list, diff, lint collision checks, and canonical ordering.
- Keep env-constrained bootstrap entries out of the unconditional digest floor.
- Bump the kind test node image for NRI adjustment validation.

## Focus areas for review

- Exact-map semantics: observed-empty versus unavailable evidence — missing evidence fails `exact` and `deny`, never passes vacuously.
- Env variants preserved through workload matching and admission history.
- NRI validation/start ordering: `exact` and `deny` reject deferred CDI device adjustments, whose env effects are unavailable at validation time.

`go build ./...` and the full `go test ./...` pass on the rebased branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,8 +199,8 @@ jobs:
       # kind is not on the runner image; pin the release binary by sha256.
       - name: Install kind
         run: |
-          curl -fsSL -o kind https://kind.sigs.k8s.io/dl/v0.30.0/kind-linux-amd64
-          echo "517ab7fc89ddeed5fa65abf71530d90648d9638ef0c4cde22c2c11f8097b8889  kind" | sha256sum -c -
+          curl -fsSL -o kind https://kind.sigs.k8s.io/dl/v0.33.0/kind-linux-amd64
+          echo "aee6151561422756b764a4ae28e7f44cda5af5a9eead3cc9985112b1de8d8e0d  kind" | sha256sum -c -
           sudo install kind /usr/local/bin/kind
 
       - name: Run the cluster integration harness
--- a/internal/admissionhistory/history.go
+++ b/internal/admissionhistory/history.go
@@ -3,6 +3,7 @@ package admissionhistory
 
 import (
 	"fmt"
+	"github.com/confidential-dot-ai/c8s/pkg/allowlist"
 	"slices"
 
 	"github.com/confidential-dot-ai/c8s/pkg/workloadclaims"
@@ -17,7 +18,7 @@ type History struct {
 
 // Record adds an admission; only a resolved record for the same ID clears an
 // unresolved container. History owns its copy of argv.
-func (h *History) Record(id, digest string, argv []string) {
+func (h *History) Record(id, digest string, argv []string, env ...*allowlist.EnvObservation) {
 	if h.byKey == nil {
 		h.byKey = map[string]workloadclaims.SandboxContainer{}
 		h.unresolved = map[string]struct{}{}
@@ -28,6 +29,9 @@ func (h *History) Record(id, digest string, argv []string) {
 	}
 	delete(h.unresolved, id)
 	c := workloadclaims.SandboxContainer{Digest: digest, Argv: slices.Clone(argv)}
+	if len(env) > 0 {
+		c.Env = env[0].Clone()
+	}
 	h.byKey[c.Key()] = c
 }
 
@@ -42,6 +46,7 @@ func (h History) Snapshot() ([]string, []workloadclaims.SandboxContainer, error)
 	for _, c := range h.byKey {
 		digests = append(digests, c.Digest)
 		c.Argv = slices.Clone(c.Argv)
+		c.Env = c.Env.Clone()
 		containers = append(containers, c)
 	}
 	slices.Sort(digests)
--- a/internal/admissionhistory/history.go
+++ b/internal/admissionhistory/history.go
@@ -18,7 +18,7 @@ type History struct {
 
 // Record adds an admission; only a resolved record for the same ID clears an
 // unresolved container. History owns its copy of argv.
-func (h *History) Record(id, digest string, argv []string, env ...*allowlist.EnvObservation) {
+func (h *History) Record(id, digest string, argv []string, env *allowlist.EnvObservation) {
 	if h.byKey == nil {
 		h.byKey = map[string]workloadclaims.SandboxContainer{}
 		h.unresolved = map[string]struct{}{}
@@ -28,10 +28,7 @@ func (h *History) Record(id, digest string, argv []string, env ...*allowlist.Env
 		return
 	}
 	delete(h.unresolved, id)
-	c := workloadclaims.SandboxContainer{Digest: digest, Argv: slices.Clone(argv)}
-	if len(env) > 0 {
-		c.Env = env[0].Clone()
-	}
+	c := workloadclaims.SandboxContainer{Digest: digest, Argv: slices.Clone(argv), Env: env.Clone()}
 	h.byKey[c.Key()] = c
 }
 
--- a/internal/allowlist/handler.go
+++ b/internal/allowlist/handler.go
@@ -87,7 +87,12 @@ func (h Handler) HandlePutWorkload(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	entry, err := pkgallowlist.ParseWorkloadJSON(body)
+	current, _, err := h.Store.LoadAll()
+	if err != nil {
+		http.Error(w, "internal server error", http.StatusInternalServerError)
+		return
+	}
+	entry, err := pkgallowlist.ParseWorkloadJSONForSchema(body, current.Schema)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusUnprocessableEntity)
 		return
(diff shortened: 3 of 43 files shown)


internal/admissionhistory/history.go
```diff
@@ -28,6 +29,9 @@ func (h *History) Record(id, digest string, argv []string) {
 	}
 	delete(h.unresolved, id)
 	c := workloadclaims.SandboxContainer{Digest: digest, Argv: slices.Clone(argv)}
+	if len(env) > 0 {
+		c.Env = env[0].Clone()
+	}
```
CluEleSsUK: I don't understand this at all - why isn't env going in the SandboxContainer like the other parts? and why only clone 0?

.github/workflows/ci.yml
```diff
@@ -223,8 +223,8 @@ jobs:
       # kind is not on the runner image; pin the release binary by sha256.
       - name: Install kind
         run: |
-          curl -fsSL -o kind https://kind.sigs.k8s.io/dl/v0.30.0/kind-linux-amd64
-          echo "517ab7fc89ddeed5fa65abf71530d90648d9638ef0c4cde22c2c11f8097b8889  kind" | sha256sum -c -
+          curl -fsSL -o kind https://kind.sigs.k8s.io/dl/v0.33.0/kind-linux-amd64
+          echo "aee6151561422756b764a4ae28e7f44cda5af5a9eead3cc9985112b1de8d8e0d  kind" | sha256sum -c -
```
CluEleSsUK: why is updating kind necessary as part of this PR?

internal/allowlist/handler.go
```diff
@@ -87,7 +87,12 @@ func (h Handler) HandlePutWorkload(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	entry, err := pkgallowlist.ParseWorkloadJSON(body)
+	current, _, err := h.Store.LoadAll()
+	if err != nil {
+		http.Error(w, "internal server error", http.StatusInternalServerError)
+		return
+	}
+	entry, err := pkgallowlist.ParseWorkloadJSONForSchema(body, current.Schema)
```
CluEleSsUK: if this is refactor is for backwards compat we should nuke it - new clusters will start afresh

internal/allowlist/store.go
```diff
@@ -40,6 +40,8 @@ const (
 var ErrInvalidWorkload = errors.New("invalid workload entry")
 
 const initSQL = `
+CREATE TABLE IF NOT EXISTS allowlist_format (id INTEGER PRIMARY KEY CHECK(id=1), schema TEXT NOT NULL);
+INSERT OR IGNORE INTO allowlist_format VALUES (1, 'c8s.allowlist/v1');
```
CluEleSsUK: don't bother with backwards compat

internal/cmds/allowlist/derive.go
```diff
@@ -128,6 +135,45 @@ deliberately absent from the derived entry.`,
...
+					return err
+				}
+			} else if envMode != allowlist.PolicyAny && envMode != allowlist.PolicyDeny {
+				return fmt.Errorf("--env must be any or deny; use --env-file for exact values")
+			}
+			used := map[string]bool{}
+			for _, part := range []struct {
+				templates  []templateContainer
+				containers []allowlist.Container
+			}{{spec.InitContainers, initContainers}, {spec.Containers, containers}} {
```
CluEleSsUK: would be clearer to have a named struct for this

internal/cmds/allowlist/diff.go
```diff
@@ -40,18 +40,22 @@ func (e entryDiff) empty() bool {
 
 // allowlistDiff is the entry-level diff of two allowlists.
 type allowlistDiff struct {
+	Schema           *changedEntry        `json:"schema,omitempty"`
```
CluEleSsUK: seems like an out of place change

internal/cmds/allowlist/lint.go
```diff
@@ -172,7 +172,7 @@ func lintOffline(al *pkgallowlist.Allowlist) []finding {
 				entriesByDigest[d] = map[string]bool{}
 			}
 			entriesByDigest[d][name] = true
-			if c.AnyArgv() {
+			if c.AnyArgv() && (c.Env.Policy == pkgallowlist.PolicyAny || c.Env.Policy == "") && c.Mounts.Policy != pkgallowlist.PolicyExact {
```
CluEleSsUK: this type of check ought to be extracted behind a meaningfully function or variable name, otherwise it's hard to read in the context of a bigger function

internal/cmds/allowlist/lint.go
```diff
@@ -245,7 +250,7 @@ func shadowFinding(wide, narrow string) finding {
 func shadows(wide, narrow pkgallowlist.Workload) bool {
 	unconstrained := map[string]bool{}
 	for _, c := range allContainers(wide) {
-		if c.AnyArgv() && c.Mounts.Policy != pkgallowlist.PolicyExact && c.Env.Policy != pkgallowlist.PolicyExact {
+		if c.AnyArgv() && c.Mounts.Policy != pkgallowlist.PolicyExact && (c.Env.Policy == pkgallowlist.PolicyAny || c.Env.Policy == "") {
```
CluEleSsUK: same here

pkg/allowlist/allowlist.go
```diff
@@ -36,6 +36,9 @@ import (
 // canonical form.
 const Schema = "c8s.allowlist/v1"
 
+// SchemaV2 adds exact environment values and requires explicit env policies.
+const SchemaV2 = "c8s.allowlist/v2"
```
CluEleSsUK: let's just keep this v1 for now until we hit `v1.0.0`

pkg/allowlist/env.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func validEnvPair(name, value string) bool {
+	return name != "" && !strings.ContainsAny(name, "=\x00") && !strings.ContainsRune(value, 0) && utf8.ValidString(name) && utf8.ValidString(value)
+}
+
+// The encoding is the UTF-8 domain string "c8s.env/v1", a NUL, then Go
+// encoding/json's compact object (sorted keys, HTML escaping enabled). Empty
+// maps encode as {}, never null. Version this encoding if it ever changes.
+func fingerprintEnv(values map[string]string) *EnvObservation {
```
CluEleSsUK: is it really necessary to fingerprint things? I think comparing key/value should be just fine

pkg/allowlist/match.go
```diff
@@ -5,20 +5,20 @@ import "fmt"
...
-// plugin gates images on a node CVM and fills Digest and Argv only.
+// Missing Env is unavailable evidence, and fails v2 exact/deny policies.
+// Legacy v1 mount/name-set policies retain their historical subset semantics.
+// Kata fills bind mounts and legacy names; both backends report v2 env evidence.
 type RunningContainer struct {
 	Digest     string
 	Argv       []string
 	BindMounts []string
-	EnvNames   []string
+	EnvNames   []string // legacy v1 only
```
CluEleSsUK: no legacy paths

pkg/allowlist/policy.go
```diff
@@ -149,3 +149,23 @@ func equalStrings(a, b []string) bool {
 	}
 	return true
 }
+
+// AdmitsProcess is the preliminary NRI create-time check. It deliberately checks
+// only digest/argv; a successful result MUST be followed by AdmitsContainer on
+// the finalized OCI spec before start, and must never authorize secret release.
+func (i *Index) AdmitsProcess(r RunningContainer) bool {
```
CluEleSsUK: why not env here?

pkg/allowlist/match.go
```diff
@@ -144,14 +134,24 @@ func anyRunning(running []RunningContainer, c Container) bool {
...
+func (c Container) admitsProcess(r RunningContainer) bool {
 	if c.Digest.String() != r.Digest {
 		return false
 	}
 	rest, ok := c.Command.matchCommand(r.Argv)
 	if !ok || !c.Args.matchArgs(rest) {
 		return false
 	}
-	return c.Mounts.admits(r.BindMounts) && c.Env.admits(r.EnvNames)
+	return true
```
AnomalRoil, line 154: `return ok && c.Args.matchArgs(rest)`

AnomalRoil (approved):

PR #587, opened 2026-09-10
feat(ratls): jitter for cert renewal

CluEleSsUK:
In order to avoid flooding hardware provider KMS on cert renewal, a 20% jitter was added to cert renewals (which already happen at 50% of the cert lifetime).


joaosa (approved):

internal/cmds/getcert/run.go
```diff
@@ -446,6 +446,11 @@ func renewalInterval(cfg config, leaf *x509.Certificate, unnamedRuns int) time.D
 			delay = half
 		}
 	}
+	// Renew up to 20% early so certificates issued together do not refresh
+	// in lockstep. Keep the already-jittered unnamed fast poll and delay floor.
+	if fifth := int64(delay) / 5; fifth > 0 {
```
AmeanAsad: nit: magic num

CluEleSsUK: have made it a CLI param with a default now

PR #593, opened 2026-09-10
ci(snp-metal): pin attestation-api to the attestation-rs commit the node image was built from

b0xtch:
The attestation-api DaemonSet image floated to `:main` while the node image bakes attestation-rs at the commit pinned in `.github/build-pins.json`. A run therefore tested two attestation-rs versions in one cluster, a combination nobody ships, and the lane went red for attestation-rs reasons on days with no c8s change (the comment this replaces records enforce-VMPL0/reject-debug-policy landing on `:main`).

Pin the DaemonSet to the same commit. Read `builds.node-image.attestation_rs_ref` from `build-pins.json` in the c8s checkout the lane already has and use `attestation-api:sha-<first 7>`. This is the version-pairing principle the lane already enforces for `c8sRef`, applied to the one component it skipped.

Edges, each tested under bash by extracting the block with `inputs.aa_ref` rendered per scenario, a fake `c8s/` checkout, and `curl` stubbed for the ghcr token and manifest probe:

```
pin present, tag exists            rc=0  AATAG=sha-d868688
pin present, tag missing           rc=1  ::error::build-pins.json pins attestation-rs d868688 but ghcr has no attestation-api:sha-0000000
pin file absent (old ref)          rc=0  AATAG=main   ::warning::attestation-api unpinned
pin file present, key missing      rc=1  ::error::... has no builds.node-image.attestation_rs_ref
aa_ref override wins               rc=0  AATAG=sha-abc1234
```

A pin that points at nothing fails rather than falling back, since that is a broken pin, not a transient. Refs older than the pin file (before c8s#523, 2026-09-03) have nothing to honour, so they warn and float. The healthy path also ran once against real ghcr (the harness shell did not export the stub) and passed, so `attestation-api:sha-d868688` is live today.

What this does to "what if attestation-rs changes": the pin moves through the existing weekly `confos-pin-watch`, which rolls the measurements and runs the reproducibility gate; last cycle attestation-rs landed `d868688` on 09-04 and c8s#548 merged the bump on 09-09. attestation-rs tip is proven on this same hardware by attestation-rs's own `confidential-e2e` on every push, so that question is answered there rather than by floating here.

Not covered: this file is vendored into confidential-ci; the re-vendor follows. The Azure lanes do not plumb `aa_ref` and take the chart default, a separate change. The c8s lane now lags attestation-rs by up to a week plus merge time; a c8s change that needs newer attestation-rs must land the bump first, the same discipline already in place for confos.

Falsified by: a run whose `resolved image tags` line shows `attestation-api=main` on a c8s ref that carries `build-pins.json`.

### DIFF
--- a/.github/workflows/snp-metal-e2e.yml
+++ b/.github/workflows/snp-metal-e2e.yml
@@ -128,14 +128,34 @@ jobs:
           # never tracked and get-cert dies with "caller cgroup names no tracked
           # container". Pin it like cds and nri already are.
           OPDIG=$(resolve_digest c8s-operator "$OPTAG")
-          # attestation-api is versioned by the attestation-rs repo, NOT c8s — a
-          # c8s-ancestry walk can never find its tags (learned the hard way:
-          # InvalidImageName, run 29473241754). :main is the canonical pointer the
-          # c8s-integration fleet uses; accepted race until c8s pins a digest.
-          # aa_ref lets a caller PIN it (tag or sha-<short>) — floating :main drifts
-          # (e.g. attestation-rs enforce-VMPL0/reject-debug-policy landed on :main and
-          # broke this dev-sandbox CVM while the boot image was unchanged).
-          AATAG="${{ inputs.aa_ref || 'main' }}"
+          # attestation-api is versioned by the attestation-rs repo, NOT c8s, so a
+          # c8s-ancestry walk can never find its tags (InvalidImageName, run
+          # 29473241754). Pin it to the attestation-rs commit the node image was
+          # built from (.github/build-pins.json, kept current by confos-pin-watch),
+          # so the DaemonSet and the copy baked into the guest are the same code.
+          # Floating :main ran two attestation-rs versions in one cluster and went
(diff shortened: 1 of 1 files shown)


b0xtch:
Withdrawing: wrong layer.

The chart already pins attestation-api by digest (`c8sComponents` -> `attestationApi.image` -> `pinnedDigest`), honoured by `c8s install --resolve-digests`. The real defect is that c8s holds two pins for the same dependency, moved by two processes: `build-pins.json` (the source commit the node image bakes, bumped weekly by pin-watch) and `pinnedDigest` (the DaemonSet image, bumped by hand per its own comment). Measured today they disagree: build-pins says `d868688` -> `sha256:fb0d98a3`, the chart pins `sha256:53a1bbe`. Three lanes deploy the stale one; snp-metal bypasses the pin entirely.

This PR would have added a third decider in one lane. The fix belongs in c8s: one pin, one mover (pin-watch stamps both), a CI check that they agree, and the lane consuming the chart's pin rather than overriding it. Follow-up PRs to come.

PR #600, opened 2026-09-11
feat(kata): removed pod-as-cvm construction

CluEleSsUK:


joaosa (approved):
🔥

node-guest-image/README.md
CluEleSsUK, line 1: this was extracted from a file that was previously kata only

PR #604, opened 2026-09-11
docs: env and mount allowlist policy is not enforced in node mode

aamirrasheed:
## Summary

Under `--cvm-mode=node`, c8s does not enforce the `mounts` or `env`
allowlist fields. The host NRI image-policy plugin is the only
container-start enforcer in this mode. Its allowlist check supplies only the
image digest and effective argv to the matcher.

An `exact` mount or environment policy therefore does not narrow node-mode
admission. A container that passes digest and argv policy also passes these
unobserved constraints.

This PR documents the behavior and its security effect. It does not change
runtime behavior.

## Enforcement path

The NRI plugin checks this observation in
`internal/cmds/nri-image-policy/plugin.go`:

```go
allowlist.RunningContainer{Digest: digest, Argv: argv}
```

`BindMounts` and `EnvNames` remain nil. In `pkg/allowlist/match.go`,
`everyIn` returns true for an empty observation. Therefore, an exact policy
cannot reject an extra mount or environment name in node mode.

Digest and argv enforcement still applies. This gap does not admit an
arbitrary image or command.

The in-guest `policy-monitor` used by `--cvm-mode=pod` reads the guest OCI
specification and supplies all four observations. It enforces `mounts` and
`env` in that mode.

CDS is a separate allowlist enforcement point. At certificate issuance it
checks digest membership and can derive a matched-workload identity from
digest and argv. It receives no mount or environment observation. A relying
party can verify the TEE evidence, certificate chain, workload name, and
allowlist document, but it cannot detect a mount or environment change from
those values.

## Security effect

A credential that can patch a Deployment can add environment variables or
Restricted-safe volume mounts to an otherwise admitted container. If it can
also supply executable content, it can use mechanisms such as `LD_PRELOAD`
with a mounted library. The image digest and argv remain unchanged.

With the default `hostNamespacePolicy` enabled, the
`deny-host-namespaces` `ValidatingAdmissionPolicy` blocks `hostPath` in
non-exempt workload namespaces. It does not block ConfigMap, Secret,
emptyDir, PVC, projected, and other Restricted-safe volume sources.

Known node-mode consumers include `confidential-inference` and
`confidential-inference-internal`. Their current checked-in production and
integration-staging allowlists use `policy: any` for `mounts` and `env`.
This is therefore a missing guarantee in their node-mode threat model, not a
bypass of an active exact constraint.

## Requirements for a runtime fix

NRI `api.Container` includes `Env` and `Mounts`. Supplying their environment
names and bind-mount destinations to `allowlist.RunningContainer` would cover
the original CRI observation.

That change alone would not prove the final container state. NRI collects
adjustments from all mutating plugins after it calls their `CreateContainer`
handlers. A later adjustment can add a mount or environment variable. c8s
also adds its inventory socket mount to selected injected sidecars after the
current image check.

A complete fix must validate the effective state after applicable NRI
adjustments, or prevent later unvalidated changes to these fields. NRI
provides `ValidateContainerAdjustment` for validation of the combined
adjustments. Exact policies must also include kubelet-added mounts and
Kubernetes-injected environment names.

`c8s allowlist lint` should continue to warn for node mode until the effective
mount and environment state is enforced.

## Verification

Review these paths in order:

1. `internal/cmds/nri-image-policy/plugin.go`: the `RunningContainer` literal
   supplies only digest and argv.
2. `pkg/allowlist/match.go`: `everyIn` accepts an empty observation.
3. `internal/cmds/policymonitor/run.go`: the pod-mode monitor supplies digest,
   argv, bind mounts, and environment names.
4. `internal/cmds/allowlist/lint.go`: node mode warns about exact mount and
   environment policy.
5. `internal/cmds/cds/attest.go` and `pkg/ratls/matchedworkload.go`: the CDS
   matched-workload path has digest and argv data, but no mount or environment
   data.

`git diff --check` and `gofmt -l internal/cmds/nri-image-policy/plugin.go`
report no issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -563,6 +563,15 @@ than let you discover them:
   *together* — every running image must be allowlisted, but no gate requires
   the set in one pod to match a single workload entry.
 
+- **`mounts` and `env` allowlist policy is not enforced under
+  `--cvm-mode=node`.** The only enforcer running there, the host NRI
+  image-policy plugin, observes digest and argv only, so an `exact` `mounts`
+  or `env` policy is vacuously satisfied and admits every container. A leaf
+  certificate and workload receipt still derive from digest and argv only, so
+  a bind mount or environment variable added at the Deployment stays
+  invisible to a relying party. See [docs/allowlist-and-capabilities.md, "Node
+  mode: not enforced"](docs/allowlist-and-capabilities.md#node-mode-not-enforced).
+
 - **Secrets and encrypted volumes under pod-as-CVM carry weaker guarantees
   than on a node.** Both work — the injected fetchers redeem their sandbox
   token from the in-guest `policy-monitor` over loopback, and `volumed
--- a/README.md
+++ b/README.md
@@ -564,13 +564,14 @@ than let you discover them:
   the set in one pod to match a single workload entry.
 
 - **`mounts` and `env` allowlist policy is not enforced under
-  `--cvm-mode=node`.** The only enforcer running there, the host NRI
-  image-policy plugin, observes digest and argv only, so an `exact` `mounts`
-  or `env` policy is vacuously satisfied and admits every container. A leaf
-  certificate and workload receipt still derive from digest and argv only, so
-  a bind mount or environment variable added at the Deployment stays
-  invisible to a relying party. See [docs/allowlist-and-capabilities.md, "Node
-  mode: not enforced"](docs/allowlist-and-capabilities.md#node-mode-not-enforced).
+  `--cvm-mode=node`.** The host NRI image-policy plugin is the only
+  container-start enforcer in this mode. Its allowlist check supplies digest
+  and argv only. An `exact` `mounts` or `env` policy therefore does not narrow
+  admission: a container that passes digest and argv policy also passes these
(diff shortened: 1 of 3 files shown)


AnomalRoil:
Closing since this has been solved by now:

 - Env not enforced: fixed by the merged https://github.com/confidential-dot-ai/c8s/pull/584. Env values are pinned and checked on the persisted spec at StartContainer, with ValidateContainerAdjustment covering later plugin adjustments, which is exactly your "validate effective state after adjustments" requirement.
  - Mounts not enforced: fixed by https://github.com/confidential-dot-ai/c8s/pull/610 (open). Bind mounts are classified by source and checked in the same final phase, including kubelet-added platform mounts.
  - lint should keep warning until enforced: #610 removes that warning since both fields are now observed.
  - CDS identity has no env observation: #584 carries env fingerprints through admission history and CDS matching.
  - hostPath and privilege side note: https://github.com/confidential-dot-ai/c8s/pull/608 covers it with the sandbox policy, though it ships in audit mode.
  - CDS probably won't be binding identity with mounts, that doesn't seem necessary to achieve the security guarantees we need.

PR #606, opened 2026-09-14
chore: rename tls-lb to router

CluEleSsUK:


AnomalRoil (changes requested):
Please also remove mention of TLS load balancer:
``` 
test/integration/run.sh
2:# Integration test for the TLS load balancer.

README.md
177:# Install the platform (node-as-CVM) and point the bundled TLS load balancer

docs/operator.md
338:confirm CDS — or the load balancer — is a genuine TEE running the expected code

internal/helmchart/c8s/templates/router-deployment.yaml
66:        # this a tightened pin would not roll the load balancer.
```

CluEleSsUK:
updated!

AnomalRoil (approved):

PR #607, opened 2026-09-14
chore: rename node to bare-metal

CluEleSsUK:
after we removed pod-as-cvm, the term 'node' is ambiguous as it applies to all c8s constructions. Renaming it bare-metal to separate it from the cloud offerings.


AnomalRoil (approved):
lgtm

joaosa (approved):

PR #609, opened 2026-09-14
feat(node-image): seal post-start exec at the runc boundary

AnomalRoil:
## Why

The kubelet runs `exec` liveness/readiness/startup probes and `postStart`/`preStop` hooks through CRI `ExecSync`, and `kubectl exec`/`cp` reach the same runtime call. `enable-debugging-handlers=false` closes only the kubelet's HTTP API: none of the other paths passes through it, and none fires an NRI hook. On a measured node image, anyone who can put argv in a probe field of an admitted pod gets code execution inside a running container and reads the output back from pod events.

Separately, nothing proved that no container was running when the image-policy plugin registered. The plugin is pre-registered in the image's containerd config and `required_plugins` blocks creation while it is absent, but the ordering that makes that true is asserted by systemd, not by the enforcer.

## Description

**`c8s-runc`, a measured runtime wrapper** (`internal/cmds/c8srunc`, `cmd/c8s-runc`). containerd's runc shim runs the binary named by the runtime handler's `BinaryName` for every verb, so a wrapper there sees every request that starts a process in a running container. A locked build denies the `exec` verb and `execve`s the real runc for every other verb with the original argv, environment and file descriptors. The posture is a build-time constant — no mode file, no allowlist, nothing on a running node changes it. Global options are parsed only to find the subcommand, modelled on the runc the pinned RKE2 ships; an option the wrapper does not recognise is a denial, because a token modelled wrongly shifts which argument the subcommand is. The denial is also written to runc's `--log` in `--log-format`, so containerd surfaces it in the pod event instead of an empty runtime error. The node image installs it as an argv alias of the `c8s` binary, dispatched by basename before cobra sees runc's argv.

**Image wiring.** `mkosi.sync` installs the wrapper at `/usr/local/bin/c8s-runc`; the baked drop-in `config-v3.toml.d/10-c8s-runc.toml` points the runc handler's `BinaryName` at it. A drop-in rather than a template line: RKE2's base template already defines that table and containerd's TOML parser rejects a duplicate; containerd merges each import over the main config key by key. `C8S_DEV=1` renders a later drop-in restoring the unwrapped runtime, from the same variant switch that re-enables the kubelet debugging handlers — the two images already have different measurements.

**Build gate.** `node-guest-image/c8s/containerdcheck` renders the profile's template the way RKE2 does (its base template vendored and pinned to the RKE2 release), merges the drop-ins containerd imports, and asserts every enabled ordinary-pod handler runs the wrapper — failing on an empty `BinaryName`, an alternate runc, a later drop-in that unwraps the handler, an unaudited shim, or a `default_runtime_name` that is not wrapped. The invariants gate adds what a rendered config cannot show: the wrapper path the build installs, the real-runtime path the wrapper is compiled with, the absence of any runtime binary RKE2's auto-detection would add a handler for (`crun`, the nvidia runtimes, the wasm shims), and the base-template pin against `RKE2_VERSION`. `go-toml` moves from an indirect to a direct dependency; it is what containerd parses the same file with.

The vendored base template follows the k3s revision behind RKE2 v1.36.4. Since k3s b51167a9 that base emits the `config-v3.toml.d/*.toml` import itself, so the baked `config-v3.toml.tmpl` no longer prepends its own `imports` line: the gate rendered the old template as a duplicate TOML key.

**Chart.** c8s-owned exec probes and the one lifecycle exec hook are replaced, since neither can pass on such a node:

- `attest-proxy --health-addr` serves `GET /healthz` by making the same round-trip over its own socket the healthcheck subcommand makes; readiness and liveness use it.
- `ratls-mesh iptables-sync --ready-addr` serves `GET /readyz` on node loopback; the startup probe that gates the mesh proxy reads it.
- The mesh cleanup sidecar runs `iptables-cleanup --keep-guard --on-shutdown`, which idles until SIGTERM and then cleans up, in place of the `preStop` hook. Native sidecars stop in reverse init order, so the ordering is unchanged; `ratlsMesh.iptablesCleanup.preStopSleepSeconds` now renders `--settle-delay`.
- `acme --ready-port` serves `GET /readyz`, reporting the same condition nginx's `probe-file` exec probe reported.

New values: `attestationApi.proxy.healthPort` (8401), `ratlsMesh.ports.iptablesReady` (15022), and the chart-internal `router.acmeReadyPort` helper (8403). A chart test walks every rendered manifest for exec probes and lifecycle exec hooks.

**Image-policy plugin.** New `policy.fatal_existing` (with `policy.boot_marker_path`, validated to require fail-closed mode, `enforce_existing`, and an absolute tmpfs path): a container present at the plugin's first registration after boot has already run, so stopping it proves nothing — the plugin logs every container it was handed and powers the node off. A plugin restart is a different event, since containerd stays up and its containers legitimately predate the new process, so a boot-scoped marker separates the two: a restart re-checks what is running through the existing startup check and escalates only there. Not while the first pull after that restart is outstanding — the plugin is Ready on the boot floor alone then, and every workload the allowlist service admitted would look unknown; that case keeps the existing stop. If the power-off syscall is refused the plugin exits, which leaves `required_plugins` blocking every new container. The setting is off by default, so the chart's containerised plugin is unchanged; the node image's baked config turns it on.

## Focus areas for review

- `parseGlobals` in `internal/cmds/c8srunc/c8srunc.go`: the flag model is taken from the runc version the pinned RKE2 ships, and the deny-on-unknown-option rule is what keeps a future runc flag from turning into a pass-through.
- The wrapper's real-runtime path is RKE2's extracted runc under its `bin` symlink. RKE2 keeps that directory on containerd's `PATH`, so `runc` still resolves by name for a process already running as root on the node; no enabled handler uses it. `docs/node-exec-mode.md` "Known limits" states this.
- The dev drop-in is rendered into a directory that also exists in the profile's `mkosi.extra`; extra-tree precedence for a shared directory is worth confirming on the first dev build.
- Whether powering the node off is the right action for a container found at first registration, and whether the "no escalation before the first allowlist pull" carve-out is drawn in the right place.
- Probe substitutions: `/healthz` and `/readyz` are reachable from other pods (the mesh one only from node loopback). They report liveness or file existence and carry no evidence.

## Not covered here

- Reporting the exec posture from a verified image manifest, so a relying party can tell a locked image from a debug one without comparing measurements: follow-up.
- The image cannot be built in this environment, so the installed binary, the rendered dev drop-in and the config RKE2 writes at boot are unverified outside the fixtures.
- `runc checkpoint` and `restore` are not aliases of `exec` and pass through; they deserve their own audit.

Related: https://github.com/confidential-dot-ai/c8s/pull/584, https://github.com/confidential-dot-ai/c8s/pull/585


internal/cmds/acme/cmd.go
CluEleSsUK, line 1: not strictly necessary, but we could add a `healthz` here too just for consistency

internal/cmds/attestproxy/cmd.go
```diff
@@ -122,20 +125,62 @@ func serve(ctx context.Context, cfg config, proxy http.Handler, listener net.Lis
 	}
 	go cmdsutil.ShutdownOnDone(ctx, srv, 5*time.Second)
 
+	// Started after the socket is bound so a probe never passes before the
+	// front door is up.
+	if cfg.healthAddr != "" {
```
CluEleSsUK, line 130: this (and readyz) feels like code that could be extracted into a reusable helper

internal/cmds/c8srunc/c8srunc.go
```diff
@@ -0,0 +1,231 @@
...
+
+// DenyMessage is the stable text a denied exec reports. Tests and the e2e
+// matrix match on it; keep it greppable in kubelet events and containerd logs.
+const DenyMessage = "c8s-runc: exec is denied on this node: the image is built in locked exec mode"
+
+// Build-time configuration, set with -ldflags -X. Defaults are the production
+// (locked) posture, so an unflagged build is never the permissive one.
+var (
+	mode     = ModeLocked
+	realRunc = "/var/lib/rancher/rke2/bin/runc"
```
CluEleSsUK, line 55: we support kubeadm too and it uses different paths

internal/cmds/nri-image-policy/bootgate.go
```diff
@@ -0,0 +1,182 @@
...
+		return true
+	}
+	if len(ctrs) == 0 {
+		g.logger.Info("first registration since boot: no container exists yet, as a static boot requires")
+		return false
+	}
+	for _, ctr := range logLines(ctx, p, pods, ctrs) {
+		g.logger.Error("container running before admission was in place", ctr...)
+	}
+	g.fatal("containers were running when the image-policy plugin first registered")
```
CluEleSsUK, line 98: this is surprising to me - shouldn't it just kill the containers? or is the fear that they've modified the host? how will this work on hosted k8s where there will be things running beforehand? do they all have to be on the allowlist?

node-guest-image/c8s/containerdcheck/main.go
```diff
@@ -0,0 +1,299 @@
...
+func (e *extraRuntimes) Set(v string) error {
+	name, binary, ok := strings.Cut(v, "=")
+	if !ok || name == "" {
+		return fmt.Errorf("want NAME=BINARY, got %q", v)
+	}
+	if *e == nil {
+		*e = extraRuntimes{}
+	}
+	(*e)[name] = binary
+	return nil
```
CluEleSsUK, line 109: jesus christ haha

node-guest-image/c8s/containerdcheck/main.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+// effectiveConfig renders the template in configDir and merges its drop-ins,
+// returning what containerd would hold after loading the imports.
+func effectiveConfig(configDir string, extra extraRuntimes) (map[string]any, error) {
```
CluEleSsUK, line 114: btw i'm fixing a bug with this after recent bump - seems containerd change their config format a bunch fuuuu

node-guest-image/c8s/containerdcheck/main.go
```diff
@@ -0,0 +1,299 @@
...
+
+// renderData mirrors k3s's templates.ContainerdConfig for a c8s node: the
+// fields the base template reads, with the values RKE2 fills in on this image.
+// Only ExtraRuntimes varies, and only for the negative tests.
+func renderData(extra extraRuntimes) map[string]any {
+	runtimes := map[string]any{}
+	for name, binary := range extra {
+		runtimes[name] = map[string]any{"RuntimeType": runcShim, "BinaryName": binary}
+	}
+	return map[string]any{
```
CluEleSsUK, line 201: i get the feeling we could codify this in a struct (or perhaps there is one we can import directly from containerd?) and use toml un/marshalling to handle a lot of the code in this file

internal/cmds/nri-image-policy/bootgate.go
```diff
@@ -0,0 +1,186 @@
...
+		imageRef := ctr.GetAnnotations()[annotationImageName]
+		line := []any{
+			"container", ctr.GetName(),
+			"image", imageRef,
+			"digest", p.resolveDigest(ctx, imageRef),
+		}
+		if pod := podByID[ctr.GetPodSandboxId()]; pod != nil {
+			line = append(line, "namespace", pod.GetNamespace(), "pod", pod.GetName())
+		}
+		lines = append(lines, line)
```
AnomalRoil: Not sure this `[]any` if better than using a `strings.Builder`

internal/cmds/nri-image-policy/bootgate.go
```diff
@@ -0,0 +1,200 @@
...
+// admitted from a CDS allowlist is then unknown to it. Powering the node off
+// on that would turn a transient CDS outage into a dead node; stopping the
+// container, which is what an ungated plugin does, stays the answer until the
+// allowlist is the real one. With no pull configured the floor is the real one.
+func (p *plugin) escalateOnRestart() bool {
+	if p.boot == nil || !p.bootRestart.Load() {
+		return false
+	}
+	if !p.cfg.PullEnabled() {
+		return true
```
CluEleSsUK:
```suggestion
	if p.cfg.PullEnabled() {
		return false
```

node-guest-image/c8s/containerdcheck/main.go
```diff
@@ -0,0 +1,318 @@
...
+		return 1
+	}
+	fmt.Fprintf(stdout, "every ordinary-pod handler runs %s: %s\n", *wrapper, strings.Join(handlers, ", "))
+	return 0
+}
+
+// extraRuntimes collects repeated -extra-runtime NAME=BINARY flags.
+type extraRuntimes map[string]string
+
+func (e extraRuntimes) String() string { return strings.Join(slices.Sorted(maps.Keys(e)), ",") }
```
CluEleSsUK, line 98: i've noticed LLMs love sticking funcs on one line. I'm surprised the linter allows it, but imo multi-line is always easier to read

CluEleSsUK (approved):

joaosa (approved):

internal/cmds/c8srunc/c8srunc.go
joaosa: yap we should include that

PR #610, opened 2026-09-14
feat(nri-image-policy): enforce bind-mount policy on a node CVM

AnomalRoil:
## Why

An allowlist entry pins the bytes and what runs them, but says nothing about what the node lays *over* those bytes at start-up. The NRI plugin observed digest, argv and — with https://github.com/confidential-dot-ai/c8s/pull/584 — the environment, but never the mount table, so a `mounts` policy was vacuously satisfied wherever that plugin was the enforcer.

A volume bound over `/etc/ld.so.preload`, `/etc/ld.so.conf.d/`, an interpreter's site directory or a dotfile the shell reads runs operator-supplied bytes as code inside an image that was reviewed, while every digest still reports as admitted. Client-side verification does not catch it: the pod keeps its genuine identity.

## Description

**Observation.** The plugin classifies each bind mount by the path the node staged its source at — the kubelet's per-pod directory and volume-plugin directories, containerd's per-sandbox directory — into `platform`, `emptyDir`, `data` (operator-supplied content) or `host`. Anything it cannot attribute is `host`. Non-bind entries (`proc`, `sysfs`, `tmpfs`, `devpts`, `mqueue`, `cgroup`) are dropped.

The check runs in the final phase https://github.com/confidential-dot-ai/c8s/pull/584 introduced, at `StartContainer`, where the spec is the persisted one. A mount table observed at `CreateContainer` precedes other NRI plugins and CDI injection, so it is not the one the container starts with. The preliminary phase stays digest and argv only.

**Schema.** `MountPolicy` gains two optional fields:

- `sandboxed` (bool) — applies the rule below to an `exact` policy.
- `reviews` (destination → string) — why bytes arriving at that destination cannot name code.

```json
"mounts": {
  "policy": "exact",
  "sandboxed": true,
  "destinations": ["/mnt/c8s-data/config", "/var/cache/nginx"],
  "reviews": { "/mnt/c8s-data/config": "yaml the app parses; loads no modules" }
}
```

Under `sandboxed`: platform mounts are admitted without an entry; an `emptyDir` needs a listed destination and nothing else; operator-supplied content is admitted only at a listed destination under `/mnt/c8s-data/` that carries a review; a host path is refused. One destination prefix is cheaper to reason about than a denylist of loader paths, since no executable, library, loader configuration or interpreter module path lives beneath it.

Both fields are optional and the marker is a field rather than a fourth `policy` value, so a node image that predates them keeps parsing the served document and applies plain destination containment. Without the marker an `exact` policy means exactly what it meant before. The chart's own entries carry no `mounts` policy at all, so a fresh `helm install` is unaffected; moving the chart's ConfigMap volumes under the prefix is separate work.

**Enforcement gaps, each named where it attaches.** hostPath is refused outright, including for measured platform containers; pinning a host source to a node-TCB rule attaches at `floorPinsHostMount` in `internal/cmds/nri-image-policy/mounts.go`. The kubelet root is the constant `/var/lib/kubelet` (RKE2's default; the node image sets no `root-dir` kubelet-arg); a node that moved it would classify every mount as `host` and refuse every sandboxed entry, which is fail-closed and bounded to entries carrying the marker.

`docs/allowlist-and-capabilities.md` regains a `Mount policy (mounts)` section — https://github.com/confidential-dot-ai/c8s/pull/584 kept only the environment one — covering plain containment and the sandboxed rule.

**lint.** Two new errors: a review outside the prefix, and a prefix destination with no review. A third refuses an `exact` `PATH`, `LD_LIBRARY_PATH`, `PYTHONPATH` or `NODE_PATH` naming a directory at or under the prefix — a sealed search path reaching it would let the reviewed exact command resolve to a file the operator wrote. `--cvm-mode` is deprecated and inert (still accepted, so an existing invocation does not fail): every policy field is now observed by the enforcer, so it silences nothing.

## Focus areas for review

- `classifyMount` in `internal/cmds/nri-image-policy/mounts.go` — the source shapes are taken from a live RKE2 node. Anything misattributed as `platform` is admitted without an entry, so that branch is the sharp one: it requires both a node-staged source *and* one of the five fixed platform destinations, and the serviceaccount projection only at its own destination.
- Phase placement: mounts are read only when `final` is set, so `AdmitsProcess` at create is unchanged.
- `MountPolicy.admits` in `pkg/allowlist/match.go` — the fallback to plain containment when the marker is absent or the observation is unclassified, and that unknown classes fall to the refusing default.
- Whether deprecating `lint --cvm-mode` rather than removing it is the right call.


docs/allowlist-and-capabilities.md
```diff
@@ -181,6 +182,95 @@ The NRI plugin enforces env after cumulative NRI adjustments, and the admission
...
+```json
+"mounts": { "policy": "exact", "destinations": ["/etc/hosts", "/config"] }
+```
+
+`exact` requires every observed bind destination to appear in `destinations`:
+what the pod's `volumeMounts` declare, plus the handful the node always adds
+(`/etc/hosts`, `/etc/hostname`, `/etc/resolv.conf`, `/dev/termination-log`,
+`/dev/shm`, the serviceaccount token). `any` is the default when the field is
+absent — unlike argv, which defaults to `deny`, because a container always
+carries a mount table it never declared and a `deny` default would refuse every
```
CluEleSsUK, line 208: if a complete `deny` is de facto impossible, `deny` semantics should be `[/etc/hosts, /etc/hostname...] and no more`, and we should default deny.

docs/allowlist-and-capabilities.md
```diff
@@ -181,6 +182,95 @@ The NRI plugin enforces env after cumulative NRI adjustments, and the admission
...
+real pod. That makes the field opt-in: a digest with no policy is constrained
+exactly as much as it was before.
+
+### Sandboxed mounts (node-as-CVM)
```
CluEleSsUK, line 212: everythign is node-as-cvm now

docs/allowlist-and-capabilities.md
```diff
@@ -181,6 +182,95 @@ The NRI plugin enforces env after cumulative NRI adjustments, and the admission
...
+| data | every other kubelet volume plugin — configMap, secret, projected, CSI, local — and volume subpaths | destination must be listed, lie under `/mnt/c8s-data/`, and carry a `reviews` string |
+| host | a hostPath volume, or any source the enforcer cannot attribute | refused |
+
+The data prefix is the point of the rule, and it is what takes the destinations
+above off the table. At a listed destination the bytes are untrusted application
+input, and one prefix is cheaper to reason about than a denylist of loader
+paths: no executable, library, loader configuration or interpreter module path
+lives under `/mnt/c8s-data/`.
+
+The `reviews` string says *why* bytes at that destination cannot name code — no
```
CluEleSsUK, line 241: don't love this tbh - clutters things up a bit

docs/allowlist-and-capabilities.md
```diff
@@ -181,6 +182,95 @@ The NRI plugin enforces env after cumulative NRI adjustments, and the admission
...
+### Sandboxed mounts (node-as-CVM)
+
+Destination containment alone still trusts the operator's choice of *what*
+lands at a destination. Add `"sandboxed": true` to an `exact` policy and the
+node enforcer classifies each bind mount by the path the node staged its source
+at, then applies one rule per class:
+
+```json
+"mounts": {
+  "policy": "exact",
```
CluEleSsUK, line 221: something not captured in this layout: volume type. In principle:
- empty in-memory storage is nearly always fine (sharing between containers in pod)
- encrypted storage is fine
- plaintext storage is probably never fine

docs/allowlist-and-capabilities.md
```diff
@@ -181,6 +182,95 @@ The NRI plugin enforces env after cumulative NRI adjustments, and the admission
...
+- **hostPath is refused outright**, including for c8s's own measured platform
+  containers. Pinning a host source to a node-TCB rule attaches at
+  `floorPinsHostMount` in `internal/cmds/nri-image-policy/mounts.go`; nothing is
+  floor until that marker lands.
+- **The kubelet root is the constant `/var/lib/kubelet`** (RKE2's default; the
+  node image sets no `root-dir` kubelet-arg). A node that moved it would
+  classify every mount as host and refuse every sandboxed entry — fail-closed,
+  and bounded by the fact that only entries carrying the marker are affected.
+
+The marker is a field rather than a fourth `policy` value so a node image that
```
CluEleSsUK, line 259: shouldn't bother with backwards compat

docs/allowlist-and-capabilities.md
```diff
@@ -181,6 +182,95 @@ The NRI plugin enforces env after cumulative NRI adjustments, and the admission
...
+predates it keeps parsing the served document — it ignores the field and applies
+plain containment — instead of failing every pull. An enforcer that reports
+destinations without saying who staged them gets that same containment check.
+
+A sealed `PATH`, `LD_LIBRARY_PATH`, `PYTHONPATH` or `NODE_PATH` that names a
+directory under the prefix undoes all of it: the reviewed exact command would
+then resolve to a file the operator wrote. `lint` refuses that too.
+
+The c8s chart's own pods mount ConfigMaps and host paths outside the prefix
```
CluEleSsUK, line 268: we could add these to the floor in https://github.com/confidential-dot-ai/c8s/pull/585

internal/cmds/allowlist/lint.go
```diff
@@ -70,7 +70,11 @@ same.`,
 	}
 	cmd.Flags().BoolVar(&online, "online", false, "also check each digest exists in its registry via crane")
 	cmd.Flags().BoolVar(&strict, "strict", false, "exit non-zero if there are any warnings")
-	cmd.Flags().StringVar(&cvmMode, "cvm-mode", "", "deployment mode the allowlist targets (pod, node, gke, aks); pod silences the mount/env scope warning")
+	// Inert since the NRI plugin began observing the mount table and the
+	// environment: nothing is left that only a guest enforcer could see. Kept
+	// accepted so an existing CI invocation does not fail on an unknown flag.
+	cmd.Flags().StringVar(&cvmMode, "cvm-mode", "", "deprecated and ignored")
```
CluEleSsUK, line 76: so we don't need this for az vs gke vs bare metal?

internal/cmds/allowlist/lint.go
```diff
@@ -271,25 +277,79 @@ func shadows(wide, narrow pkgallowlist.Workload) bool {
...
+// content it was carved out for.
+func mountPolicyFindings(al *pkgallowlist.Allowlist) []finding {
+	var out []finding
 	for _, name := range slices.Sorted(maps.Keys(al.Workloads)) {
 		for _, c := range allContainers(al.Workloads[name]) {
-			var fields []string
-			if c.Mounts.Policy == pkgallowlist.PolicyExact {
-				fields = append(fields, "mounts")
+			m := c.Mounts
+			if !m.Sandboxed {
```
CluEleSsUK, line 293: I still haven't fully grasped what sandboxing is in this context

internal/cmds/nri-image-policy/plugin.go
```diff
@@ -356,10 +356,14 @@ func (p *plugin) checkImage(ctx context.Context, cfg *config, namespace, podName
...
-	return p.checkImagePhase(ctx, cfg, namespace, podName, containerName, imageRef, argv, observed, true)
+	return p.checkImagePhase(ctx, cfg, namespace, podName, containerName, imageRef, argv, observed, nil, true)
 }
 
-func (p *plugin) checkImagePhase(ctx context.Context, cfg *config, namespace, podName, containerName, imageRef string, argv []string, observed *allowlist.EnvObservation, final bool) (imageVerdict, string) {
+// mounts is the container's classified bind-mount table (observeMounts). Like
+// the environment it is read only in the final phase, where the spec is the
+// persisted one: CreateContainer runs before other plugins and CDI injection,
+// so a mount table observed there is not the one the container starts with.
+func (p *plugin) checkImagePhase(ctx context.Context, cfg *config, namespace, podName, containerName, imageRef string, argv []string, observed *allowlist.EnvObservation, mounts []allowlist.ObservedMount, final bool) (imageVerdict, string) {
```
CluEleSsUK, line 369: don't love boolean parameters

pkg/allowlist/allowlist.go
```diff
@@ -96,10 +96,35 @@ type ArgvPolicy struct {
...
+// them falls back to plain containment, so a served document carrying the
+// marker stays enforceable on a node image that predates it.
 type MountPolicy struct {
 	Policy       string   `json:"policy"`
 	Destinations []string `json:"destinations,omitempty"`
-}
+	// Sandboxed opts an exact policy into the classification above. It is a
+	// separate field rather than a fourth policy value so a consumer that
+	// predates it keeps parsing the document instead of failing every pull.
+	Sandboxed bool `json:"sandboxed,omitempty"`
```
CluEleSsUK: ah now I understand it. Again, we should just drop this and only support newest format

pkg/allowlist/match.go
```diff
@@ -151,12 +190,56 @@ func (c Container) admitsProcess(r RunningContainer) bool {
...
+// containment, which is what the field meant before Sandboxed existed.
+func (p MountPolicy) admits(r RunningContainer) bool {
 	if p.Policy != PolicyExact {
 		return true
 	}
-	return everyIn(destinations, p.Destinations)
+	if !p.Sandboxed || r.Mounts == nil {
+		return everyIn(observedDestinations(r), p.Destinations)
+	}
+	for _, m := range r.Mounts {
```
CluEleSsUK, line 206: i'm sure there is a funny func thing for this

CluEleSsUK (changes requested):

CluEleSsUK:
Closing in favour of stacked PRs

PR #552, opened 2026-09-07
refactor(e2e): test signed launches and baked Kubernetes workloads

joaosa:
Update the hardware lanes to generate signed launch bundles, bootstrap through
the authenticated credential service, and test the image's baked Kubernetes
workloads. TDX resolves and stages the image published for the selected source
commit; SNP verifies the staged image's paired source and launch metadata.
Automatic TDX acceptance validates the publication's signed-launch evidence.

The readiness check waits for CDS, router, mesh and operator rollout, checks
their required read-only public policy mounts, and verifies the CDS front door
against the full server image/key policy. This follows #664's migration away
from duplicate host application services and the runtime ConfigMap.

## Stack

#662 → #663 → #664 → #552

Land the complete stack together before publishing or deploying the new node
images: guest boot and hardware launchers share the signed launch contract.

## Validation

- Workflow regression tests passed after restacking onto #664.
- TDX publication acceptance harness: 47 assertions passed.
- Readiness script checked against the actual rendered workloads with a stubbed
  API/CLI: valid state plus seven failure cases passed.
- Shellcheck and actionlint passed with the repository's existing exclusion for
  the intentional remote-shell expression.
- Coverage passed on retry after the unchanged base's ACME test hit a local
  port collision. TDX/SNP builds and reproducibility comparison passed.
- TDX/SNP hardware launches were not rerun locally.

## Original review feedback

The original review remains on this PR. Its feedback is mapped to the owning
changes in #662 (policy inputs and bootstrap), #663 (launch CLI and identity
consumer), and #664 (baked manifests and inventory clients), with links back
to each original thread. Shared nonce/challenge extraction remains deferred
as requested by the reviewer.

CI still reports the advisory mutation timeout; the Meat waiter propagates
that canceled workflow status. The functional checks above passed.

PR #611, opened 2026-09-14
feat(node-image): bound the operator credential and add a logs-only role

joaosa:
Stacked on #552 (`node-image/baked-chart`).

## Summary

Enable `kubectl logs` on the locked node image, and stop issuing cluster-admin from cred-release so the baked guards hold against every released credential. Adds a logs-only credential for people who should not hold the operator role.

### Node image
- Kubelet debugging handlers stay on: they back `kubectl logs` and have no logs-only switch. `exec`, `attach`, `port-forward` and ephemeral containers are closed in admission for every principal by the baked `confos-pod-exec` policy. `C8S_DEV=1` skips that AddOn through an RKE2 `.skip` marker rendered by `mkosi.sync`.
- `c8s:node-operators` is bound to a new wildcard-free `c8s-node-operator` ClusterRole instead of `cluster-admin`. It manages tenant workloads, namespaces, namespaced RBAC, ConfidentialWorkloads and network policy cluster-wide and reads the rest. No `bind`/`escalate`/`impersonate`, no cluster-scoped RBAC or admission writes, no `nodes/proxy`, `pods/proxy` or `services/proxy` (these reach the kubelet's exec endpoint without a `pods/exec` admission check), no PV/StorageClass/HelmChart/Node/CSR-approval writes, no `podsecurityexemptions` grant.
- New `confos-operator-scope` policy denies the same set in admission for any `c8s:*` group regardless of RBAC, plus every write in the PodSecurity-exempt namespaces. It short-circuits for every other principal.
- New `c8s-log-reader` ClusterRole bound to `c8s:log-readers`: pods, `pods/log`, namespaces and events, read-only.
- `psa-ready.sh` waits for all new objects, proves the operator-scope deny and allow paths with server-side dry-runs as a synthetic principal carrying the operator group, and compares every live guard object against a reference copy that `mkosi.sync` stages under `/usr/lib/confai/guards` on the read-only root. `server/manifests` is on the writable overlay because RKE2 stages its bundled charts there; a shadowed or edited guard now keeps `:8443` closed.

### cred-release / get-kubeconfig
- The release request carries a `role`. Empty or `operator` keeps the current identity; `log-reader` issues `CN=log-reader, O=c8s:log-readers` with a 24h default TTL. Unknown roles are a 400, never a fallback. The role sits in the token-bound body, so it cannot be swapped in transit.
- New `--log-cert-org`, `--log-cert-cn`, `--log-cert-ttl` flags, spelled out in the baked unit like the operator flags.
- `c8s get-kubeconfig --role log-reader`.

### What a released credential can no longer do
Lower a namespace below restricted, exec, write in `kube-system` or `local-path-storage`, mount the host through a PersistentVolume or a `nodePath` StorageClass, or edit or delete any guard or binding. Root-equivalent access to the guest exists only through the in-guest `rke2.yaml`.

## Validation
- `internal/cmds/credrelease`, `internal/cmds/getkubeconfig`, `internal/helmchart` Go tests pass. New coverage: per-role Subject and TTL, unknown-role rejection, incomplete role config, role on the wire, role allowlist checks (rejects wildcards, dangerous verbs, guarded resources), both new policies' shape, CEL evaluation of the operator-scope policy across 36 request shapes.
- `node-guest-image/tests/psa-ready-test.sh`: 42 cases pass (13 new for guard comparison, 8 for operator scope, 3 for the pod-exec dev skip).
- The new invariants block passes in isolation; the full script needs a confos checkout.
- The four e2e scripts the metal lanes run with the released kubeconfig were audited against the bounded role and fit. They have not been run on hardware with this image.

## Notes
- The image measurement changes; regenerate the manifest on the next build.
- Immutability of the guards against guest root is not claimed. The reference on the verity root plus the gate comparison is the strongest available, since RKE2 must write `server/manifests`.


internal/cmds/credrelease/handler.go
```diff
@@ -21,14 +23,57 @@ const maxBodyBytes = 1 << 16 // 64 KiB
...
+// Roles the operator may request. The operator key authorizes every release,
+// so the role is a choice of how much of the operator's own authority the
+// issued cert carries, not a second gate: RoleOperator manages tenant
+// workloads cluster-wide (the node image binds it to a bounded role that
+// cannot reach the baked guards, never cluster-admin), RoleLogReader reads
+// pods and their logs and nothing else. Each maps to a
+// distinct cert Subject (group + user) the cluster's RBAC binds.
+const (
+	RoleOperator  = "operator"
+	RoleLogReader = "log-reader"
```
CluEleSsUK, line 35: in the OIDC design, I'd envisaged that we would define capabilities in c8s and orgs would define roles outside in some other system. 
That said, I see how it makes sense to group different capabilities into roles within c8s too.
Maybe a good middle ground is:
- c8s itself assesses capabilities
- c8s cli has a notion of roles and can create operator-key-signed capability grants that fan out roles -> capabilities

This design would dovetail nicely with future OIDC work and would allow issuing long-lived `log-reader` grants to people

CluEleSsUK: also from a versioning perspective, we don't have any compat issues if we add/remove capabilities from roles

node-guest-image/c8s/mkosi.sync
```diff
@@ -104,14 +100,35 @@ printf '[Service]\nEnvironment=CRED_PLATFORM=%s\n' "$PLATFORM" \
...
+# Reference copies of the guard AddOns on the read-only root. RKE2 must be
+# able to write server/manifests (it stages its bundled charts there), so
+# that directory lives on the writable overlay and a root process could
+# shadow a guard. psa-ready.sh waits for every guard staged here and compares
+# the live objects against the copy before cred-release serves; a drifted
+# guard keeps :8443 closed. GUARDS is the single statement of what is a
+# guard (node-guest-image-invariants.sh reads it); the profile tree is the
+# single source of each: this copies, never edits.
+#
+# C8S_DEV=1 builds skip the baked pod-exec-policy AddOn so `kubectl exec`/
```
CluEleSsUK, line 112: so the debug marker in k8s doesn't _also_ turn on exec/attach?

node-guest-image/README.md
```diff
@@ -27,10 +27,12 @@ Layout:
...
   here. Point `CONFOS_DIR` at a confos checkout (default: a sibling dir).
-  The locked image runs the kubelet with `enable-debugging-handlers=false`,
-  so `kubectl exec`, `attach`, `port-forward`, and `logs` fail for every
-  kubeconfig holder; `C8S_DEV=1` turns them back on (with the serial
-  autologin), at a different measurement.
+  The locked image keeps the kubelet debugging handlers on so `kubectl logs`
+  works for every kubeconfig holder, and bakes the `pod-exec-policy.yaml`
+  AddOn so `kubectl exec`, `attach`, `port-forward` and `debug` (ephemeral
+  containers) are denied in admission for everyone, the operator included;
+  `C8S_DEV=1` skips that AddOn (with the serial autologin), at a different
```
CluEleSsUK, line 34: ah okay, my question is answered

node-guest-image/c8s/mkosi.sync
joaosa: Yes. A dev build (`C8S_DEV=1`) leaves out the rule that blocks exec, attach and port-forward, so those work again for debugging. The normal build always includes that rule.

Logs keep working in both builds. The part of the kubelet that serves logs is the same part that serves exec, so we cannot switch one off without the other; instead we block exec at the point where the API server checks requests. A dev build has a different measurement, so it cannot be passed off as a normal one.

internal/cmds/credrelease/handler.go
joaosa: Agreed on the direction. For this PR I kept two fixed roles on purpose: it is the smallest change that stops handing out cluster-admin and lets someone read logs without being an operator.

Splitting roles into individual permissions, with the command line issuing signed grants that turn a role into a list of permissions, is the right next step and fits the OIDC work. Tracked as a follow-up. As you say, changing what a role includes would then not break anyone holding an older grant.

CluEleSsUK (approved):

PR #614, opened 2026-09-14
feat(allowlist): carry mount evidence through workload matching

CluEleSsUK:
Record classified bind mounts in the sandbox inventory and include them in the identity of each admitted container. Pass that evidence to certificate matching and secret release, so an exact mount policy cannot be satisfied by an inventory that never observed the mount table.

Add --mounts-file to allowlist derive for per-container mount policies, and name the preliminary and final admission phases explicitly.

The default mount policy and persistent-volume rules remain open decisions; this commit does not complete the PR's mount-policy redesign.


internal/cmds/nri-image-policy/mount_storage.go
```diff
@@ -0,0 +1,61 @@
...
+	slaves, err := os.ReadDir(filepath.Join(device, "slaves"))
+	if err != nil {
+		return false
+	}
+	for _, slave := range slaves {
+		if hasC8sCryptDevice(filepath.Join(device, "slaves", slave.Name()), seen) {
+			return true
+		}
+	}
+	return false
```
CluEleSsUK: this could be simplified with `slices.ContainsFunc`

internal/cmds/nri-image-policy/mounts.go
```diff
@@ -132,7 +141,42 @@ func kubeletVolumePlugin(source string) (string, bool) {
...
-	return strings.HasPrefix(source, kubeletRoot+"/pods/") || strings.Contains(source, sandboxSegment)
+	if strings.HasPrefix(source, kubeletRoot+"/pods/") {
+		return true
+	}
+	for _, root := range containerdRoots {
+		if strings.HasPrefix(source, root+sandboxSegment) {
+			return true
+		}
+	}
+	return false
```
CluEleSsUK: could use `slices.ContainsFunc`

internal/cmds/nri-image-policy/mounts.go
```diff
@@ -132,7 +141,42 @@ func kubeletVolumePlugin(source string) (string, bool) {
...
+	case "/etc/hosts":
+		return podSource && podPath == "etc-hosts"
+	case "/dev/termination-log":
+		return podSource && strings.HasPrefix(podPath, "containers/") && strings.Count(podPath, "/") == 2
+	case "/etc/hostname":
+		return nodeStaged(source) && strings.Contains(source, sandboxSegment) && strings.HasSuffix(source, "/hostname")
+	case "/etc/resolv.conf":
+		return nodeStaged(source) && strings.Contains(source, sandboxSegment) && strings.HasSuffix(source, "/resolv.conf")
+	case "/dev/shm":
+		return nodeStaged(source) && strings.Contains(source, sandboxSegment) && strings.HasSuffix(source, "/shm")
```
CluEleSsUK: why not compare the full suffixes here? e.g. `/etc/hostname`

internal/helmchart/c8s/templates/validations.yaml
```diff
@@ -65,7 +65,7 @@
...
 {{- $img := get $c "image" -}}
 {{- if and (get $c "enabled") (not (get $c "cdsExempt")) (get $img "digest") (not (hasKey $covered (get $img "digest"))) -}}
-{{- fail (printf "VALIDATION_ERROR kind=uncovered_component_digest: nriImagePolicy.policy.mode=fail-closed with deriveComponents=false, but %s is digest-pinned (%s) and no nriImagePolicy.bootstrapAllowlist.workloads entry admits that digest under any command and args. The plugin would deny it on its own node. Add such an entry (c8s allowlist add writes the same shape), set deriveComponents=true, or use c8s install --resolve-digests." (get $c "name") (get $img "digest")) -}}
+{{- fail (printf "VALIDATION_ERROR kind=uncovered_component_digest: nriImagePolicy.policy.mode=fail-closed with deriveComponents=false, but %s is digest-pinned (%s) and no nriImagePolicy.bootstrapAllowlist.workloads entry admits that digest under unrestricted command, args, mounts and env policies. The plugin would deny it on its own node. Add such an entry (c8s allowlist add writes the same shape), set deriveComponents=true, or use c8s install --resolve-digests." (get $c "name") (get $img "digest")) -}}
```
CluEleSsUK: there are some changes on main that apply here

pkg/allowlist/match.go
```diff
@@ -3,34 +3,38 @@ package allowlist
...
+	Source      string       `json:"-"`
+	Class       MountClass   `json:"class"`
+	Storage     MountStorage `json:"storage"`
 }
 
+type MountStorage string
+
+const (
+	MountMemory    MountStorage = "memory"
+	MountEncrypted MountStorage = "encrypted"
```
CluEleSsUK: our base image is configured to create an ephemeral encryption key at boot for use with encrypted mounts that lie outside the CDS secret brokering flow. Are these also covered by "encrypted"?

PR #624, opened 2026-09-15
refactor(allowlist): unify container policy constraints

CluEleSsUK:
Container policy behavior was spread across normalization, preliminary admission, final admission, linting, and secret-release eligibility. Adding mount provenance to each path would duplicate the same field inventory again.

This introduces private typed constraint adapters for process, mounts, and environment. Command and args form one process constraint because command consumes the effective argv prefix before args evaluates the remainder. The allowlist remains a concrete, statically typed document; no interfaces or dynamic maps enter the JSON/YAML schema.

This PR is intentionally behavior-preserving and is the base for the replacement complete mount-pinning PR.

Validation:

- `GOTOOLCHAIN=go1.26.3 go test ./...`
- `git diff --check`


internal/cmds/allowlist/lint.go
```diff
@@ -422,6 +422,5 @@ func isTagForm(image string) bool {
 }
 
 func hasUnconstrainedRuntimePolicy(c pkgallowlist.Container) bool {
-	return c.AnyArgv() && c.Mounts.Policy != pkgallowlist.PolicyExact &&
-		(c.Env.Policy == pkgallowlist.PolicyAny || c.Env.Policy == "")
+	return c.Unconstrained()
```
CluEleSsUK: this seems like an anaemic function - we can just replace the callsite with `c.Unconstrained()`. I'd suggest the func is renamed `IsUnconstrained`

pkg/allowlist/allowlist.go
```diff
@@ -213,15 +213,27 @@ func (w Workload) Digests() []types.Digest {
 // AnyArgv reports whether the container is admitted whatever it runs: command
 // and args both any.
 func (c Container) AnyArgv() bool {
-	return c.Command.Policy == PolicyAny && c.Args.Policy == PolicyAny
+	return (processConstraint{command: &c.Command, args: &c.Args}).unconstrained()
```
CluEleSsUK: it might make sense for `processConstraint` etc to be members created in the constructor rather than on demand

pkg/allowlist/mountenv_test.go
CluEleSsUK: are these all the tests required for this new functionality? seems like we should have more scenarios

PR #625, opened 2026-09-15
feat(allowlist): pin mount provenance end to end

CluEleSsUK:
## Summary

Enforce typed mount policies consistently at NRI admission and during workload matching for certificate stamps and secret release. Builds on the typed constraint model merged in #624.

Previously, mount rules listed destinations only, omitted mount policies meant `any`, and inventory did not carry mount evidence into release matching. This change checks the destination, source class, and backing storage together. For example, a declared data mount on verified encrypted storage can match; the same destination backed by an unknown host path cannot.

## Policy behavior

| Policy | Behavior |
| --- | --- |
| `deny` (also the default when omitted) | Admit only verified runtime-provided platform mounts. |
| `exact` | Require the observed non-platform mounts to match the declared destination/kind rules, with memory or verified encrypted storage. |
| `any` | Explicitly leave mounts unconstrained. |

Exact rules support two kinds:

- **`emptyDir`**: a volume belonging to the current pod, backed by memory or verified encrypted storage.
- **`data`**: operator-supplied data, including ConfigMap, Secret, projected, PVC/CSI, subpath, and reserved `c8s-volume-*` mounts. Destinations must be below `/mnt/c8s-data/`. Encrypted-volume placeholders count as data even before volume propagation.

Platform mounts such as `/etc/hosts`, `/etc/resolv.conf`, and `/dev/shm` form an implicit baseline: their expected source must belong to the current pod or sandbox, and their destination must match. Unrecognized sources are classified as `host`; constrained policies refuse them. Missing evidence, duplicate non-platform destinations, class mismatches, and unknown storage also fail closed.

Example container mount policy:

```json
{
  "policy": "exact",
  "rules": [
    {"destination": "/var/cache/app", "kind": "emptyDir"},
    {"destination": "/mnt/c8s-data/config", "kind": "data"}
  ]
}
```

## Implementation

- **NRI observation:** handlers classify platform, emptyDir, kubelet data, subpath, and unknown mounts, binding runtime source paths to the current pod/sandbox.
- **Storage evidence:** Linux distinguishes tmpfs, verified encrypted device mappings, and unknown storage. Overlay inspection selects the most specific containing mount, so a nested mount cannot inherit an ancestor overlay's storage evidence.
- **Scratch provenance:** require a boot-local record matching the current boot ID, device number, `scratch` mapper name, CRYPT UUID, and backing serial `confai-scratch`. That serial depends on the `confidential-os-builder` initrd contract. The guest scratch-enforcement script publishes the record before RKE2 starts.
- **Inventory and release:** retain each distinct `(digest, argv, env, mounts)` admission for the sandbox lifetime and pass mount evidence through CDS, secret release, and diagnostics. Source paths remain node-local and are excluded from JSON. `mounts: []` means observed empty; `null` or missing means unavailable. No separate observation flag is needed.
- **Platform bootstrap:** generated installer, uninstall, containerd-prep, and local-path helper entries explicitly allow their required host mounts while retaining command/args pins. Generated platform component entries explicitly use `mounts: any`; pause containers retain `deny`. Operator mount constraints remain intact and prevent promotion into the unconstrained boot base. Integration fixtures declare their intended mount permissions, and admission diagnostics include mount-policy failures.
- **CLI and documentation:** `allowlist derive --mounts-file` accepts a container-name map covering every init and main container. Documentation includes policy examples, source/storage behavior, and inventory semantics.
- **Constraint integration:** admission continues to evaluate all typed constraints. Lint uses `IsUnconstrained()` across process, environment, and mounts; secret-grant eligibility retains the existing `ArgvPinned()` requirement, with env/mount constraints checked during release matching.

## Compatibility and limits

- This changes mount-policy defaults and replaces destination-only rules with typed rules. Existing mount policies must be updated; no compatibility adapter or schema-version migration is included.
- Rules constrain mount destination, class, and storage protection; they do not hash or pin mounted file contents.
- Non-Linux storage inspection reports `unknown`.
- `derive` still reads a Kubernetes object as JSON; it does not automatically inspect a running cluster or render Helm charts. The mounts file supplies policy intent, and the node verifies runtime evidence.

## Validation

- Full race-enabled Go suite: `GOTOOLCHAIN=go1.26.3 C8S_REQUIRE_IPTABLES_TRANSLATE=1 make test C8S_TEST_COUNT=`.
- CI-pinned golangci-lint v2.12.2, formatting, CRD consistency, and diff checks.
- Scratch-enforcement shell tests (11 cases) and Darwin ARM64 CLI compilation.
- CI on `a95c6966`: Build & Test, Lint, Integration, Integration (cluster), Coverage gate, and commit signatures all pass.
- Regression coverage for unavailable versus empty evidence across JSON and release matching, mount identity/deduplication, pod/sandbox ownership, nested mounts, incomplete scratch provenance, encrypted-volume placeholders, full admission of rendered platform installer mounts, and preservation of operator mount constraints in bootstrap generation.

## Related work

Supersedes the mount work in #610 and #614; both remain open.


CluEleSsUK:
Closing in favour of stack

PR #628, opened 2026-09-16
feat(nri): observe mount provenance and storage protection

CluEleSsUK:
## Summary

Classify bind mounts by their source ownership and backing storage so admission can distinguish verified platform mounts, pod-owned emptyDir volumes, operator data, and unknown host paths. This is the observation foundation for mount pinning; admission and release policy behavior remain unchanged in this PR.

For example, `/etc/hosts` is a platform mount only when its source belongs to the current pod. A path belonging to another pod is classified as host-controlled even when its destination looks legitimate.

## Changes

- Add typed mount observations and handlers for platform, emptyDir, kubelet data, subpath, and unknown mounts. Bind source paths to the current pod and sandbox identities.
- Inspect Linux storage for tmpfs, encrypted scratch, and c8s crypt/verity mappings. Select the most specific containing mount when following overlay upper directories.
- Publish and validate scratch provenance against the current boot ID, device number, mapper name, CRYPT UUID, and backing-device serial. The `confai-scratch` serial follows the confidential-os-builder initrd contract.
- Add ownership, encrypted-volume placeholder, nested-overlay, malformed-provenance, crypt/verity, and cyclic-device regression tests.

## Boundaries

- The observer is introduced here and connected to admission in the next PR.
- Unknown or incomplete storage proof yields `unknown`; non-Linux inspection also yields `unknown`.
- Source paths are diagnostic node-local data and are excluded from JSON. Observations establish provenance and storage protection, not hashes of mounted file contents.
- Scratch enforcement now requires complete provenance before the node starts.

## Validation

- Full race-enabled Go suite: `GOTOOLCHAIN=go1.26.3 C8S_REQUIRE_IPTABLES_TRANSLATE=1 make test C8S_TEST_COUNT=`.
- golangci-lint v2.12.2: no issues.
- Scratch-enforcement shell suite: 11 passing cases.
- `git diff --check`.

## Stack

1. [#628 — Observe mount provenance and storage protection](https://github.com/confidential-dot-ai/c8s/pull/628) → `main`
2. [#629 — Enforce mount policies end to end](https://github.com/confidential-dot-ai/c8s/pull/629) → #628
3. [#630 — Add mount policy authoring and lint checks](https://github.com/confidential-dot-ai/c8s/pull/630) → #629

Part 1 of 3, based on `main`. Split from #625. Followed by mount-policy enforcement, then CLI authoring and lint support. Commits are intentionally unsigned drafts for the author to sign before merge. Cluster integration results must come from CI.


pkg/allowlist/match.go
```diff
@@ -14,6 +14,15 @@ type RunningContainer struct {
 	Env        *EnvObservation
 }
 
+// ObservedMount is a bind mount classified by the node. Source is diagnostic
+// node-local detail and is excluded from workload identity serialization.
+type ObservedMount struct {
+	Destination string       `json:"destination"`
+	Source      string       `json:"-"`
```
AnomalRoil, line 21: Is this really true? I don't see any special treatement of the serialisation of `ObservedMount` and its `Source`

internal/cmds/nri-image-policy/mount_storage_linux.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+	seen[clean] = true
+	var fs unix.Statfs_t
+	if err := unix.Statfs(clean, &fs); err != nil {
+		return allowlist.MountUnknown
+	}
+	if fs.Type == unix.TMPFS_MAGIC {
+		return allowlist.MountMemory
+	}
+	if uint64(fs.Type) == overlayFSMagic {
```
AnomalRoil:
```suggestion
const overlayFSMagic int64 = 0x794c7630

// scratchDiskSerial must match the scratch disk serial expected by
// confidential-os-builder's initrd; changes require coordination with that repo.
const scratchDiskSerial = "confai-scratch"

type linuxStorageInspector struct {
	sysDevBlock    string
	sysClassBlock  string
	mountInfo      string
	provenanceFile string
	bootIDFile     string
}

func newMountStorageInspector() mountStorageInspector {
	return linuxStorageInspector{
		sysDevBlock: "/sys/dev/block", sysClassBlock: "/sys/class/block",
		mountInfo: "/proc/self/mountinfo", provenanceFile: "/run/c8s/scratch-provenance.json",
		bootIDFile: "/proc/sys/kernel/random/boot_id",
	}
}

func (i linuxStorageInspector) Inspect(source string) allowlist.MountStorage {
	return i.inspect(source, map[string]bool{})
}

func (i linuxStorageInspector) inspect(source string, seen map[string]bool) allowlist.MountStorage {
	clean := filepath.Clean(source)
	if seen[clean] {
		return allowlist.MountUnknown
	}
	seen[clean] = true
	var fs unix.Statfs_t
	if err := unix.Statfs(clean, &fs); err != nil {
		return allowlist.MountUnknown
	}
	if fs.Type == unix.TMPFS_MAGIC {
		return allowlist.MountMemory
	}
	if fs.Type == overlayFSMagic {
```

AnomalRoil: you can just have the right constant type instead of doing casts in comparisons

AnomalRoil (approved):
Low-key would have liked more option-related tests, but LGTM.

pkg/allowlist/match.go
CluEleSsUK: this gets used in #629  - got added a bit early in the stack heh

CluEleSsUK:
added a few additional tests too

PR #634, opened 2026-09-16
fix(uninstall): admit the host sweep under the nri-image-policy argv pins

CluEleSsUK:
`c8s uninstall` deadlocks on node-as-CVM clusters: since #553 the served allowlist admits the `nri-image-policy` image only under the installer DaemonSet's three exact argvs, and the host sweep — same image, two different argv shapes — is refused by the baked plugin, which is still enforcing from its in-memory snapshot when the sweep runs. After this change the chart pins the sweep's argv shapes at install time, and the sweep converges: the plugin admits it under the new pins.

## Description

The plugin at sweep time can only consult the document seeded at install time (CDS, the allowlist write path, is gone with the release), so the fix is in the chart's seed: `c8s.argvPinnedEntries` gains the sweep's two container shapes — `sh -c <host-sweep.sh>` and `/bin/sleep 2147483647` — on the entry it already writes for the nri-image-policy digest. `host-sweep.sh` moves into the chart (`files/scripts/`) and the CLI reads it from the same embedded tree (`helmchart.HostSweepScript()`), so the pin and the sweep argv cannot drift apart within a version.

Across versions, the pin is rendered by the installed chart while the sweep argv came from the CLI — the same deadlock one script edit later. `c8s uninstall` therefore replays the argv the release's allowlist seed ConfigMap pins (read before `helm uninstall` deletes it), identifying the shapes by content (the script's `c8s host sweep` banner; the pause as the entry's one non-`/bin/sh -c` shape) and falling back to its embedded spec on any miss. The fallback preserves today's behavior for pre-fix releases and `--host-sweep-only`.

Two hardenings ride the new shapes (argv-exact + env-any pins the argv string, not what PATH resolves): the pause is shell-free `/bin/sleep 2147483647`, and the script pins its own PATH. The three pre-existing shapes share the class; hardening them is a separate, follow-up-sized change.

## Changes

- `_allowlist.tpl`: pin the host-sweep script and `/bin/sleep 2147483647` on the nri-image-policy seed entry.
- `host-sweep.sh` moves to `internal/helmchart/c8s/files/scripts/`; new `helmchart.HostSweepScript()` serves it to the CLI with the chart's exact normalization.
- `c8s uninstall` replays the pinned sweep/pause argv from the release's seed ConfigMap (banner/non-sh-c content rules, ambiguity falls back fail-closed).
- `--host-sweep-only` sweeps with the nri image at the CLI's own version tag: on a same-version baked node the cached pin admits it (busybox is denied); the retry-after-failure path the CLI's error text recommends now works.
- Pause goes shell-free; the sweep script pins `PATH` (`unset IFS ENV CDPATH` keeps POSIX default splitting for the intentional `$RESTART_COMMAND` word-split).
- Tests: `pinned_seed_sweep_test.go` asserts the seed admits both sweep argvs (default and baked renders), denies mutations, and keeps exactly one match per content rule; fake-exec tests cover replay verbatim (divergent bytes), position-independence, and every fallback cell.
- Kind integration lane: the nri digest leaves the any-argv floor and gains the chart-rendered pinned entry, so the lane's existing `c8s uninstall` exercises the production admission shape per-PR; a served-allowlist assertion proves no any-argv hole for the digest. (A `snp-metal-e2e` uninstall step for the fail-closed cell is staged as a follow-up patch — this branch could not carry workflow edits.)

## Focus areas for review

- The content-based shape selection in `seedPinnedSweepArgvs` (pkg/allowlist's parse sorts containers, so position was unusable) and its fallback matrix.
- Byte-identity chain: chart pin normalization vs `HostSweepScript()` vs the Go-marshalled DaemonSet argv — pinned by `TestChartPinnedSeedAdmitsHostSweepArgv`, which fails on the pre-fix chart.
- The kind lane's single-render restructure (render before `c8s install`, pinned entry injected into values.yaml afterwards) and the new served-allowlist assertions.
- Known limitation: clusters installed with v0.20.0 have cached seeds without the sweep shapes, so a fixed CLI still cannot sweep them — recovery is reinstall a fixed chart, then uninstall (CDS returns and the plugin re-pulls a document carrying the shapes). A plugin/containerd restart between `helm uninstall` and the sweep drops the node to the baked base floor, which carries no nri digest by construction; that residual fail-closed window is pre-existing.


CluEleSsUK:
looks like lots of code, but it's just the uninstall job

PR #642, opened 2026-09-17
feat(allowlist): support pinned host mounts

CluEleSsUK:
Host mounts currently require an unrestricted mount policy. Exact host rules now bind the source path, destination, and read-only mode through admission and release matching.

- Add exact host rules with clean absolute source paths and explicit mount-level access matching.
- Carry source commitments and access modes through inventory serialization, ordering, and admission keys.
- Keep host-backed workloads host-dependent and reject unavailable or unsupported host evidence.
- Check host mounts in search-path lint and cover source/access mismatches, serialization, and malformed bind sources.

Review focus: source commitments across the trust boundary, access-mode matching, and fail-closed handling of relative sources and recursive access options.

Validation: full race-enabled Go suite, golangci-lint (zero issues), and git diff --check passed with Go 1.26.3.

Stacked on #630. Draft: floor workload pinning remains unfinished pending actual OCI mount captures from the fixed node image; static-pod mounts depend on existing files and containerd resolves source symlinks. The floor policies are unchanged in this draft.


internal/admissionhistory/history.go
```diff
@@ -33,7 +33,16 @@ func (h *History) Record(id, digest string, argv []string, env *allowlist.EnvObs
 	delete(h.unresolved, id)
 	c := workloadclaims.SandboxContainer{Digest: digest, Argv: slices.Clone(argv), Env: env.Clone(), Mounts: slices.Clone(mounts)}
 	slices.SortFunc(c.Mounts, func(a, b allowlist.ObservedMount) int {
-		return cmp.Or(strings.Compare(a.Destination, b.Destination), strings.Compare(string(a.Class), string(b.Class)), strings.Compare(string(a.Storage), string(b.Storage)))
+		if order := cmp.Or(strings.Compare(a.Destination, b.Destination), strings.Compare(string(a.Class), string(b.Class)), strings.Compare(string(a.Storage), string(b.Storage)), strings.Compare(a.HostSourceDigest, b.HostSourceDigest)); order != 0 {
```
CluEleSsUK: this is a horror-show - let's extract to a func and make it more readable/abstract

internal/cmds/allowlist/lint.go
```diff
@@ -287,8 +287,8 @@ func searchPathFindings(al *pkgallowlist.Allowlist) []finding {
 					continue
 				}
 				for _, rule := range c.Mounts.Rules {
-					if rule.Kind == pkgallowlist.MountData && searchPathReachesMount(value, rule.Destination) {
-						out = append(out, errorf("workload %q container %s pins %s to a search path overlapping data mount %q; operator-supplied content could be loaded as code", name, c.Digest.String(), variable, rule.Destination))
+					if (rule.Kind == pkgallowlist.MountData || rule.Kind == pkgallowlist.MountHost) && searchPathReachesMount(value, rule.Destination) {
```
CluEleSsUK: this could be extracted to a meaningfully named function rather than being imperative

internal/cmds/nri-image-policy/mounts.go
```diff
@@ -64,7 +64,11 @@ func (o mountObserver) Observe(pod *api.PodSandbox, ctr *api.Container) []allowl
 	ctx := MountContext{PodUID: pod.GetUid(), SandboxID: pod.GetId(), Container: ctr, Storage: o.storage}
 	out := make([]allowlist.ObservedMount, 0, len(ctr.GetMounts()))
 	for _, m := range ctr.GetMounts() {
-		if m == nil || !path.IsAbs(m.GetSource()) { // filesystem mounts carry no host bytes
+		if m == nil {
+			continue
+		}
+		bind := m.GetType() == "bind" || slices.Contains(m.GetOptions(), "bind") || slices.Contains(m.GetOptions(), "rbind")
```
CluEleSsUK: again, extract to meaningful function name that describes the intent

internal/cmds/nri-image-policy/mounts.go
```diff
@@ -186,7 +190,19 @@ func cutTwoOrMore(s string) (string, string, bool) {
 type unknownMountHandler struct{}
 
 func (unknownMountHandler) Observe(_ MountContext, m *api.Mount) (allowlist.ObservedMount, bool) {
-	return observed(m, allowlist.MountHost, allowlist.MountUnknown), true
+	mount := observed(m, allowlist.MountHost, allowlist.MountUnknown)
```
CluEleSsUK: this function never actually returns false so it feels a bit... odd? overcomplex?

pkg/allowlist/allowlist.go
```diff
@@ -454,7 +456,14 @@ func normalizeMountRules(rules []MountRule) error {
 }
 
 func validateMountRuleKind(r MountRule) error {
+	if r.Kind != MountHost && (r.Source != "" || r.ReadOnly) {
```
CluEleSsUK: okay these sort of checks feel like they'd be served by refactoring into multiple implementations based on mount and serving each type behind an interface - we've had a bunch of switches at every leve

pkg/allowlist/mount_rule.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (emptyDirRule) admits(m ObservedMount) bool {
+	return m.Storage == MountMemory || m.Storage == MountEncrypted
+}
+
+func (emptyDirRule) hostIndependent() bool { return true }
+
+type dataRule struct{ emptyDirRule }
+
```
CluEleSsUK: format these with multiple lines - single lines are ugly

pkg/allowlist/mount_rule.go
```diff
@@ -0,0 +1,68 @@
...
+	default:
+		return nil
+	}
+}
+
+type emptyDirRule struct{ MountRule }
+
+func (r emptyDirRule) validate() error {
+	if r.Source != "" || r.ReadOnly {
+		return fmt.Errorf("mount source and readOnly require kind host")
```
CluEleSsUK: I  don't see how kind host is related/checked here

pkg/allowlist/mount_rule.go
CluEleSsUK: and apply similar formatting elsewhere

pkg/allowlist/host_mount_test.go
AnomalRoil, line 1: The two halves of the feature are tested separately (observer output in `TestHostMountSourceAndAccess`, policy matching in `pkg/allowlist/host_mount_test.go`), but nothing tests the whole end to end: an `exact` policy with a host rule evaluated against `mountObserver.Observe` output, in the style of `mount_phase_test.go`.

pkg/workloadclaims/container_key.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 // Key identifies a (digest, argv, env, mounts) admission in the
-// cumulative inventory. Mount source paths are excluded from identity.
+// cumulative inventory. Host source commitments and access modes are included in identity.
 // The framing must be injective: a collision erases historical evidence and can
 // allow a sandbox to match a workload it did not actually run.
 func (c SandboxContainer) Key() string {
```
AnomalRoil, line 14: I'm not sure I like this `Key` function and how it's returning a string and yet is using 0s and 1s or Varint encoding in bytes.

pkg/allowlist/mount_rule.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+type mountRuleBehavior interface {
+	validate() error
+	admits(ObservedMount) bool
+	hostIndependent() bool
+}
```
AnomalRoil, line 12: Please add global interface guards for your types, so that we'd catch a drift at build time

pkg/allowlist/host_mount.go
```diff
@@ -0,0 +1,17 @@
...
+	"encoding/hex"
+	"path"
+	"strings"
+)
+
+// HostSourceDigest commits to an exact clean absolute source path.
+func HostSourceDigest(source string) string {
+	if !path.IsAbs(source) || path.Clean(source) != source || strings.ContainsRune(source, 0) {
+		return ""
+	}
```
AnomalRoil: using the empty string as an error feels likely to shoot us in the foot later on

AnomalRoil (approved):
Couple of nits

pkg/allowlist/host_mount.go
CluEleSsUK: updated

pkg/allowlist/host_mount_test.go
CluEleSsUK: updated

pkg/allowlist/mount_rule.go
CluEleSsUK: never knew of this, nice idea

pkg/workloadclaims/container_key.go
CluEleSsUK: got some nils there now

PR #663, opened 2026-09-18
feat(launch): add signed node configuration bundles

joaosa:
Add `c8s node launch-config new`, `add-agent`, and `stage`, plus
`c8s keys sign-launch`. The existing `c8s launch-config` path remains available.
Bundles contain distinct server/agent keys, fresh join tokens, signed role
documents, and the complete server identity policy. Staging verifies the exact
signed bytes and measured launch-key binding before publishing role state.
Guest boot activation follows in the next layer.

This uses a different binding contract from [#546](https://github.com/confidential-dot-ai/c8s/pull/546):
the launch document is signed by the key bound into TDX RTMR3 or SNP HOSTDATA.
#546 commits a configuration manifest hash through TDX MRCONFIGID or SNP
HOSTDATA. It remains separate and is not included in this stack.

## Stack

#662 → #663 → #664 → #552

Land the complete stack together before publishing or deploying the new node
images: guest boot and hardware launchers share the signed launch contract.

## Validation

- Full Linux Go suite passed; coverage: 91.29%.
- Focused race tests and Linux Go lint passed.
- Linux amd64 slim node build passed.
- Both launch-command paths created signed bundles and added agents successfully.

## Original review feedback

- [Stack and node CLI integration](https://github.com/confidential-dot-ai/c8s/pull/552#discussion_r4046130959): the launch bundle commands are available under `c8s node launch-config`; the existing `c8s launch-config` path uses the same implementation.
- [Named measured identity](https://github.com/confidential-dot-ai/c8s/pull/552#discussion_r4046216126): launch verification consumes `MeasuredIdentity` from #662 and retains the image, key-binding and missing-register rejection checks.

## Gemini review follow-up

- [Server-key parse errors](https://github.com/confidential-dot-ai/c8s/pull/663#discussion_r4046393522) now propagate before role authorization.
- The current launch-digest access also handles [the original loader finding](https://github.com/confidential-dot-ai/c8s/pull/662#discussion_r4046393235): missing identities and empty digest lists fail before staging tokens or role markers.

PR #662, opened 2026-09-18
feat(attestation): preserve launch anchors and authenticate bootstrap

joaosa:
Attestation consumers now preserve each image tuple and its authorized launch
key together. Kubeconfig bootstrap obtains fresh, operator-authenticated
evidence over the credential service's RA-TLS endpoint before requesting a
credential. Existing images can use an explicit legacy `--attest-url`.

This foundation also makes coverage baseline checks reliable for the stack.

## Stack

#662 → #663 → #664 → #552

Land the complete stack together before publishing or deploying the new node
images: guest boot and hardware launchers share the signed launch contract.

## Validation

- Baseline validation: full Linux Go suite passed at 91.30% coverage; focused race tests and Linux Go lint passed.
- Follow-up validation: full Helm chart, NRI and shared policy-loader suites passed; installer/render-values policy tests and Linux-target lint passed.

## Original review feedback

- [Policy naming](https://github.com/confidential-dot-ai/c8s/pull/552#discussion_r4046196963), [JSON source clarity](https://github.com/confidential-dot-ai/c8s/pull/552#discussion_r4046204149) and [helper flags](https://github.com/confidential-dot-ai/c8s/pull/552#discussion_r4046257643): explicit `--image-policy-file` and `--image-policy-json` inputs distinguish complete JSON policies from legacy newline digest files. The old `--measurements-config` and `--measurements-config-json` flags are removed; callers must use the canonical image-policy flags.
- [Repeated loading logic](https://github.com/confidential-dot-ai/c8s/pull/552#discussion_r4046200262): policy-source selection and legacy input loading share helpers, including NRI pull and inventory verification; attestation-go retains parsing and verification ownership.
- [Named result instead of tuple](https://github.com/confidential-dot-ai/c8s/pull/552#discussion_r4046216126): `MeasuredIdentity` carries the operator binding result and attestation-go image identity; #663 updates the launch consumer.
- [Evidence dependency](https://github.com/confidential-dot-ai/c8s/pull/552#discussion_r4046222167) and [clock dependency](https://github.com/confidential-dot-ai/c8s/pull/552#discussion_r4046223137): evidence generation and clock access use single-method interfaces.
- [Nonce module](https://github.com/confidential-dot-ai/c8s/pull/552#discussion_r4046211747): shared nonce/challenge extraction remains follow-up work, as the reviewer explicitly requested.

## Gemini review follow-up

[...]

PR #664, opened 2026-09-18
feat(node-image): bake core Kubernetes workloads and pinned images

joaosa:
Bake the complete CDS, router/certificate helpers, RA-TLS mesh and operator
workloads into the measured node image. The embedded chart renders at build
time to `/var/lib/rancher/rke2/server/manifests/c8s-integration.yaml`; pinned
OCI archives under RKE2's image directory let every core container start with
`imagePullPolicy: Never`.

The host retains attestation, signed-launch verification/staging, RKE2/containerd,
NRI admission and attested credential release. Boot preparation merges the
measured chart seed into NRI's system floor before RKE2 starts, avoiding a CDS
bootstrap dependency. Core Pods mount only public launch inputs read-only:
peer listeners accept the authorized role keys, while CDS clients require the
server's full image/key policy. Signed TLS hostnames are read from fixed files.
The private launch directory and enrollment tokens are never mounted into Pods.

This removes the parallel systemd implementations of CDS, mesh, get-cert and
nginx, the runtime ConfigMap, and text-based YAML splitting. CDS and router
remain singletons. The docs describe role placement, token exposure, Pod/VM
restart behavior and image upgrades. Initial manifests and image content are
measured; live Kubernetes API changes remain within the existing trusted
cluster-admin boundary. Mixed-image rolling upgrades and HA remain unsupported.

## Stack

#662 → #663 → #664 → #552

Land the complete stack together before publishing or deploying the new node
images: guest boot and hardware launchers share the signed launch contract.
Published `C8S_REF` images must contain the new SAN-file flags; `C8S_BINARY`
overrides only the host binary and embedded chart.

## Validation

- Native CLI/rendering and complete Helm chart tests passed, including TDX/SNP renders.
- Linux tests for certificate handling, CDS, launch staging, node preparation,
  NRI admission and the Linux node-services command passed.
- Linux-targeted Go lint: 0 issues; shellcheck and actionlint passed.
- 62 systemd boot checks and 13 image-invariant regression assertions passed.
- Offline ORAS archive tests passed, preserving full multi-platform digests,
  canonical image names and deterministic bytes. A local containerd 2.3.5
  import/manifest/rootfs probe also passed for both fixture platforms.
- TDX/SNP image builds and their reproducibility comparison passed in CI,
  along with Build & Test, coverage, integration and cluster integration.
- TDX/SNP hardware cold boot and full RKE2 unpack were not run locally.

The 14 human review comments are addressed by the workload migration,
structured rendering, simplified host bootstrap and revised lifecycle docs.

CI still reports the advisory mutation timeout; the Meat waiter propagates
that canceled workflow status. The functional checks above passed.


cmd/c8s/node_image.go
```diff
@@ -0,0 +1,235 @@
...
+func splitNodeImageArtifacts(rendered []byte) (integration, nginx, seed []byte, err error) {
+	namespace, err := namespaceManifest(nodeImageNamespace)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	var ns corev1.Namespace
+	if err := yaml.Unmarshal(namespace, &ns); err != nil {
+		return nil, nil, nil, err
+	}
+	ns.Labels["confidential.ai/baked"] = "true"
```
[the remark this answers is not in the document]
joaosa: addressed

internal/cmds/nodeservices/prepare.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/confidential-dot-ai/c8s/internal/cmds/launchconfig"
+	"github.com/confidential-dot-ai/c8s/internal/fileutil"
+	"github.com/confidential-dot-ai/c8s/pkg/allowlist"
+	"gopkg.in/yaml.v3"
+)
+
+// Prepare writes only fixed service inputs. rootDir rebases paths for tests;
+// production calls it with an empty root after authenticating launch.yaml.
+func Prepare(rootDir string, d *launchconfig.Document) error {
```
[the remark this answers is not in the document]
joaosa: addressed

internal/cmds/nodeservices/services.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(data)), nil
+}
+
+// PublishNodeIP selects the same address as RKE2, including an explicit
+// authenticated node.ip. RootDir only rebases the fixed output for tests.
+func PublishNodeIP(rootDir string, d *launchconfig.Document) error {
+	address := d.Node.IP
```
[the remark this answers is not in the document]
joaosa: addressed

cmd/c8s/node_image.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		return err
+	}
+	args = appendDistroInstallArgs(args, "rke2")
+	args = appendSingleNodeInstallArgs(args, true)
+	args = append(args,
+		"--set", "node.bakedServices=true",
+		"--set", "nriImagePolicy.enabled=false",
+		"--set", "nriImagePolicy.bootstrapAllowlist.deriveComponents=true",
+		"--set", "volumed.enabled=false",
```
CluEleSsUK, line 118: do we bake this into the image?

cmd/c8s/node_image.go
```diff
@@ -0,0 +1,238 @@
...
+		defer os.RemoveAll(tmp)
+		chartDir = filepath.Join(tmp, helmchart.ChartRoot)
+	}
+	args, err := appendCvmModeInstallArgs(nil, "bare-metal", cfg.platform)
+	if err != nil {
+		return err
+	}
+	args = appendDistroInstallArgs(args, "rke2")
+	args = appendSingleNodeInstallArgs(args, true)
+	args = append(args,
```
CluEleSsUK, line 114: not sure how you feel, but I'm in favour of rendering some literal kubernetes manifests or a helm chart and baking them into the `/var/lib/rke2` directory. Removing layers of indirection might help keep our sanity

cmd/c8s/node_image.go
```diff
@@ -0,0 +1,238 @@
...
+		ns.Labels = make(map[string]string)
+	}
+	ns.Labels["confidential.ai/baked"] = "true"
+	namespace, err = yaml.Marshal(ns)
+	if err != nil {
+		return nil, nil, nil, err
+	}
+	var manifests bytes.Buffer
+	manifests.Write(namespace)
+	r := k8syaml.NewYAMLReader(bufio.NewReader(bytes.NewReader(rendered)))
```
CluEleSsUK, line 179: doing any kind of text parsing in go makes baby jesus cry

docs/operator.md
```diff
@@ -83,13 +84,19 @@ The main source directories are:
...
 support a non-CVM install shape or a bring-your-own CDS endpoint shape.
 
-### Authenticated launch configuration
+This is the guest-activation layer of a stack. Its signed launch contract
+requires the matching hardware-launcher changes in #552. Land the complete
+stack together before publishing or deploying images from it.
+
+`c8s install` (including `--cvm-mode=bare-metal`) is for chart-managed clusters.
+The [measured node image](../node-guest-image/README.md) starts CDS, NRI,
+RA-TLS mesh and its TLS front door as baked services. Its Kubernetes operator,
```
CluEleSsUK, line 93: vs baking the yaml in, this may require a VM restart for upgrade. This could be a bug or a feature depending on how you look at it.

docs/operator.md
```diff
@@ -211,6 +217,11 @@ calls `runtimemeasure.VerifyBinding` for that same pin, so an image cannot
 borrow another entry's authorized key. c8s passes these complete pins through
 `remote.Policy.Images`; legacy flags that cannot carry anchors are refused
 where they would weaken enforcement.
+The server publishes only the CDS URL and server policy to the public
+`c8s-node-runtime` ConfigMap in `c8s-system` (`cds-url`, `cds.json`); join tokens
+and private keys do not enter that ConfigMap. The operator forwards the full
```
CluEleSsUK, line 222: would join tokens public be a problem in practice? without a ratls cert you can't contact anything on the node (or can you? 🤔 I wonder if you run non-cw pods could this be problematic?)

internal/cmds/nodeservices/services.go
```diff
@@ -0,0 +1,131 @@
...
+	case d.Role != launchconfig.Server:
+		return nil, fmt.Errorf("%s is a server-only service", service)
+	}
+	cds := "--cds-url=" + d.CDSURL()
+	api := "--attestation-api-url=" + apiURL
+	pins := "--measurements-config=" + launchDir + "cds.json"
+	switch service {
+	case "attest-proxy":
+		return []string{"attest-proxy", "--socket=/var/run/nri-image-policy/attestation-api.sock", "--socket-gid=65532", "--upstream=" + apiURL}, nil
+	case "cds":
```
CluEleSsUK, line 47: this switch seems:
a) misformatted because nested switches appear at the same distance
b) complex because we have `case service == "attest-proxy"` then `switch service .. case "attest-proxy"`

internal/cmds/nodeservices/services.go
```diff
@@ -0,0 +1,131 @@
...
+			"--measurements-config=" + launchDir + "peers.json", "--operator-keys=" + launchDir + "operator-pubkey",
+			"--max-request-size=262144", "--san-validation=false",
+			"--dns-san-pattern=^[a-z0-9-]+[.][a-z0-9-]+[.]svc$",
+			"--dns-san-pattern=^" + regexp.QuoteMeta(d.TLSSAN) + "$"}, nil
+	case "mesh":
+		return []string{"ratls-mesh", "--platform=" + d.Image.Platform, api,
+			"--kubeconfig=" + kubeletConfig, "--node-ip=" + nodeIP,
+			"--measurements-config=" + launchDir + "peers.json",
+			"--cds-measurements-config=" + launchDir + "cds.json", cds,
+			"--cert-mode=cds", "--cert-dns-san=ratls-mesh.c8s.svc",
```
CluEleSsUK, line 61: i recommend one flag per line for readability. If the consumer can take `string` instead of `[]string` we could use multiline template strings too

internal/helmchart/c8s/helpers/services/README.md
```diff
@@ -3,26 +3,35 @@
...
 the node-local socket in `nriImagePolicy.hostPaths.runtimeDir`.
-When `attestationApi.enabled=true`, the socket volume/mount helpers expose the directory at its host path with a read-only mount and
+When either the chart API or baked services are enabled, the socket volume/mount
+helpers expose the directory to chart components at its host path with a read-only mount and
 `DirectoryOrCreate` hostPath. The webhook rebases the socket path for injected
 get-cert sidecars, and nri-image-policy NRI-mounts the directory read-only into
 those sidecars.
 
-In bare-metal CVM mode with the chart API disabled, `c8s.attestationApiHostIPEnv`
+In legacy bare-metal CVM mode with both the chart API and baked services disabled, `c8s.attestationApiHostIPEnv`
```
CluEleSsUK, line 25: can nuke legacy advice

node-guest-image/c8s/mkosi.extra/etc/systemd/system/c8s-get-cert.service
CluEleSsUK, line 1: ah yeah, it's sort of annoying with guest-baked services that we have to run an out-of-cluster get-cert :/

node-guest-image/c8s/mkosi.extra/etc/systemd/system/c8s-nginx.service
```diff
@@ -0,0 +1,25 @@
+[Unit]
+Description=Measured node HTTPS front door
+Requires=rke2-role.service c8s-get-cert.service
+After=rke2-role.service c8s-get-cert.service cds-attest.service allowlist-proxy.service
+ConditionPathExists=/run/confos/role-server
```
CluEleSsUK, line 5: calling out that there is only 1 per cluster. I think we all agreed this was fine for now, but we might rethink this in future

node-guest-image/c8s/mkosi.sync
```diff
@@ -116,36 +118,110 @@ C8S_REF=$(cat "$SYNC_INPUTS/c8s-ref" 2>/dev/null || true)
...
+    if ! PATH="$helm_dir/linux-amd64:$PATH" "$C8S_TARGET" node-image render \
+        --hardware-platform "$LABEL_PLATFORM" \
+        --kube-version "${RKE2_VERSION%%+*}" \
+        --image-repository "$OPERATOR_REPO" \
+        --image-digest "$OPERATOR_DIGEST" \
+        --output-dir "$out"; then
+        echo "c8s sync: FATAL ${C8S_REF} must provide a working 'c8s node-image render'; publish this branch's images and select that C8S_REF" >&2
+        exit 1
+    fi
+    install -D -m 0644 "$out/c8s-integration.yaml" \
```
CluEleSsUK, line 219: should this be in all images? honestly I haven't reviewed this file much because it's big and the mkosi stuff is a bit of black magic to me heh

node-guest-image/README.md
```diff
@@ -104,17 +122,110 @@ endpoint. Updating only the operator CLI does not add this guest endpoint.
...
 for the explicit legacy attestation URL option and verification rules.
 
-## Signed launch tools
+## Measured services and Kubernetes integration
+
+One image contains all service binaries and role conditions:
+
+| Service | Server | Agent |
+|---|---|---|
+| Local attestation API and Unix-socket proxy | Run | Run |
```
CluEleSsUK, line 131: run vs disabled seems like a weird dichotomoy - enabled vs disabled, or oneshot vs enabled vs disabled if certain things are only a single run

node-guest-image/README.md
```diff
@@ -104,17 +122,110 @@ endpoint. Updating only the operator CLI does not add this guest endpoint.
...
+`c8s/mkosi.sync` resolves the shared c8s binary and operator container digest
+from `C8S_REF`, then runs the binary's `c8s node-image render` command in the
+build tools tree with a checksum-pinned Helm binary downloaded from the
+official Helm release. Rendering uses the pinned RKE2 Kubernetes version,
+so chart capabilities match the guest. Helm stays outside the guest image.
+The command renders the embedded chart once, preserving its Kubernetes
+operator, CRDs, RBAC, webhook and admission policies while extracting nginx
+configuration and the CDS component seed for the host.
+Static Kubernetes resources are rendered during `mkosi.sync` and baked at
+`/var/lib/rancher/rke2/server/manifests/c8s-integration.yaml`. Platform labels
```
CluEleSsUK, line 163: aha so they _do_ land here. I'm torn - I honestly find systemd units really easy to read and understand, but if we render _some_ manifests here we'd might as well render them all

node-guest-image/c8s/mkosi.extra/etc/systemd/system/c8s-nginx.service
CluEleSsUK, line 1: if we ever upgrade `router` to do more than nginx, we'll need to rethink this (e.g. for the overencrypted channel)

CluEleSsUK (commented):
worth discussing the yaml vs helm vs systemd split - imo this is the most contentious PR in the stack

cmd/c8s/node_image.go
joaosa: addressed

cmd/c8s/node_image.go
joaosa: addressed

cmd/c8s/node_image.go
joaosa: addressed

docs/operator.md
joaosa: addressed

docs/operator.md
joaosa: addressed

internal/cmds/nodeservices/services.go
joaosa: addressed

internal/cmds/nodeservices/services.go
joaosa: addressed

internal/helmchart/c8s/helpers/services/README.md
joaosa: addressed

node-guest-image/c8s/mkosi.extra/etc/systemd/system/c8s-get-cert.service
joaosa: addressed

node-guest-image/c8s/mkosi.extra/etc/systemd/system/c8s-nginx.service
joaosa: addressed

node-guest-image/c8s/mkosi.sync
joaosa: addressed

node-guest-image/README.md
joaosa: addressed

node-guest-image/README.md
joaosa: addressed

node-guest-image/c8s/mkosi.extra/etc/systemd/system/c8s-nginx.service
joaosa: addressed

joaosa:
> worth discussing the yaml vs helm vs systemd split - imo this is the most contentious PR in the stack

@CluEleSsUK Ended up going with measured built-time rendered helm here

cmd/c8s/node_image.go
joaosa: Yes. Those values only exist at build time: `mkosi.sync` runs `c8s node-image render`, and the output lands in the image at `/var/lib/rancher/rke2/server/manifests/c8s-integration.yaml`, so it is part of the measurement

cmd/c8s/node_image.go
joaosa: Ended up going with exactly that. The embedded chart renders once at build time into RKE2's `server/manifests` directory, and the parallel systemd versions of CDS, mesh, get-cert and nginx are gone, along with the runtime ConfigMap. No Helm binary, chart archive or values go into the guest

cmd/c8s/node_image.go
joaosa: The string splitting is gone. The render output is decoded with the Kubernetes YAML/JSON decoder into unstructured objects, patched (namespace label, image pins, pull policy) and re-marshalled. Still YAML, but structured rather than text.

docs/operator.md
joaosa: I guess this is a feature. Any core version change is a rebuild plus relaunch with new signed pins, and mixed-image rolling upgrades are unsupported by the launch schema. Both docs now say this explicitly. A Pod restart reuses the baked image, so only version changes need the relaunch.

docs/operator.md
joaosa: They are not public. Tokens stay in the root-only staged launch directory and never enter the public `/run/c8s-node` mount or any ConfigMap (the ConfigMap itself is gone now). My agent says:

`The underlying worry is valid though: the mesh does not wrap 9345 or 6443, so anyone holding a token, including a non-cw pod, can attempt RKE2 enrollment. What limits that: the agent token only grants agent enrollment, and neither token gives the launch signing key or passes the image-and-role-key policy, so an enrolled rogue node still gets nothing from CDS. The remaining gap is that launch media is signed but not encrypted, so an infra operator who can read the ISO or outer Secret can read tokens. Documented under "Join tokens remain secret RKE2 enrollment credentials"; moving tokens off the launch media is a follow-up.`

node-guest-image/c8s/mkosi.extra/etc/systemd/system/c8s-get-cert.service
joaosa: Agreed, and it is gone. Certificate renewal and the ACME helper now run as containers in the router Deployment, so nothing out of cluster fetches certs anymore.

node-guest-image/c8s/mkosi.extra/etc/systemd/system/c8s-nginx.service
joaosa: Added this to the docs: CDS and the router are singleton Deployments pinned to the server, with the router forced to one replica when the node is baked. Server-only placement is intentional for this round.

node-guest-image/c8s/mkosi.sync
joaosa: There is one image per hardware platform, shared by both roles, so the render runs in each platform build with the matching platform label. Agent images carry the manifests too, but only `rke2-server` applies `server/manifests`, after signed launch staging picks the role. The README calls this out as intentional.

node-guest-image/README.md
joaosa: Reworked the table into Component / Lifecycle / Placement, where lifecycle is the actual kind (systemd oneshot, systemd service, containerd plugin, DaemonSet, singleton Deployment) and placement says which role runs it.

node-guest-image/README.md
joaosa: Went with render them all. systemd keeps only what has to exist before RKE2 or outside it: the local attester, signed-launch verification, NRI admission, RKE2 itself and attested credential release. Everything application-shaped is in the chart.

node-guest-image/c8s/mkosi.extra/etc/systemd/system/c8s-nginx.service
joaosa: The nginx unit is gone, so an overencrypted channel would be a change to the router Deployment in the chart rather than to a host unit.

joaosa:
> worth discussing the yaml vs helm vs systemd split - imo this is the most contentious PR in the stack

Ended up going with measured Helm: the chart renders at build time into RKE2's manifests directory and is part of the image measurement, while systemd only keeps the pre-RKE2 host pieces. See the PR body for the full split.

cmd/c8s/node_image.go
CluEleSsUK: I actually meant `volumed`. I had a look in some of the other bits and it looks like we don't bake `volumed` in as a systemd unit nor render it here, so I think encrypted volumes wouldn't work (but it's possible I missed it somewhere)

cmd/c8s/node_image.go
```diff
@@ -0,0 +1,347 @@
...
+		}
+		var object unstructured.Unstructured
+		if err := object.UnmarshalJSON(raw.Raw); err != nil {
+			return nil, fmt.Errorf("decode chart resource: %w", err)
+		}
+		kind, name := object.GetKind(), object.GetName()
+		if object.GetAPIVersion() == "" || kind == "" || name == "" {
+			return nil, fmt.Errorf("rendered chart resource requires apiVersion, kind and metadata.name")
+		}
+		switch kind {
```
CluEleSsUK: i'd extract each of the branches here to funcs for readability

internal/cmds/getcert/run.go
```diff
@@ -151,7 +152,8 @@ alongside a workload that uses the obtained certificate.`,
 
 	_ = cmd.MarkFlagRequired("cds-url")
 	_ = cmd.MarkFlagRequired("attestation-api-url")
-	_ = cmd.MarkFlagRequired("san")
+	cmd.MarkFlagsOneRequired("san", "san-file")
+	cmd.MarkFlagsMutuallyExclusive("san", "san-file")
```
CluEleSsUK, line 156: ah nice - it's possible we could reuse this to save a bunch of checks in other spots

internal/cmds/nodeservices/prepare.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	bakedData, err := os.ReadFile(path("/usr/lib/c8s/allowlist-seed.json"))
+	if err != nil {
+		return err
+	}
+	bakedSeed, err := allowlist.ParseJSON(bakedData)
+	if err != nil {
+		return fmt.Errorf("baked workload seed: %w", err)
+	}
+	if d.Role == launchconfig.Server {
```
CluEleSsUK, line 46: for these configs, I'd prefer a) separate functions for server and agent and b) struct definitions

internal/helmchart/c8s/templates/cds.yaml
```diff
@@ -102,10 +102,15 @@ spec:
              and (with persistence) an RWO volume, so pin it to a known node.
              The CDS digest is always in the NRI floor (c8s.imageAllowlist),
              so image policy admits CDS on whichever node it lands. */}}
+      {{- if .Values.node.baked }}
```
CluEleSsUK, line 111: do baked `agent`s also get node-role control-plane?

internal/helmchart/c8s/templates/router-deployment.yaml
```diff
@@ -67,6 +70,10 @@ spec:
         checksum/measurements-config: {{ .Values.cds.measurementsConfig | sha256sum }}
         {{- end }}
     spec:
+      {{- if .Values.node.baked }}
```
CluEleSsUK, line 75: ah, maybe the rationale overriding my previous comment is that routers only get rendered on the server node, therefore if baked then control plane. Could break in an update though

node-guest-image/c8s/mkosi.extra/etc/systemd/system/attest-proxy.service
```diff
@@ -0,0 +1,21 @@
...
+[Service]
+Type=simple
+ExecStart=/usr/local/bin/c8s attest-proxy --socket=/var/run/nri-image-policy/attestation-api.sock --socket-gid=65532 --upstream=http://127.0.0.1:8400
+Restart=on-failure
+RestartSec=5
+NoNewPrivileges=yes
+ProtectHome=yes
+ProtectSystem=strict
+ReadWritePaths=/run
+CapabilityBoundingSet=CAP_CHOWN
```
CluEleSsUK, line 21: is this to chown api.sock or the /sev-guest or something else?

cmd/c8s/install.go
```diff
@@ -1348,7 +1377,7 @@ func appendDistroInstallArgs(helmArgs []string, distro string) []string {
 // `--cvm-mode` (deployment shape) and `--hardware-platform` (CPU TEE) are
 // ORTHOGONAL axes. pod/node/gke pair with either SEV-SNP
 // (--hardware-platform sev-snp, default) or Intel TDX (--hardware-platform
-// tdx). aks uses the Azure vTPM path regardless of the CPU TEE: the node's
+// tdx). aks uses the Azure vTPM path regardless of the CPU Family: the node's
```
AnomalRoil: pretty sure it's not really the CPU Family, but the CPU TEE type

cmd/c8s/node_services_linux.go
AnomalRoil, line 1: This kind of file-based build tags _linux without explicit build tags in the file should still warrant a comment as for why we need this only on Linux, IMO

AnomalRoil (commented):
lgtm

cmd/c8s/node_image.go
joaosa: yeah I think you're right. this might need another PR 🤔

cmd/c8s/node_image.go
joaosa: Addressed in fb8fbde8: the loop body is now `setNodeImageNamespace`, `collectWorkloadImages` and `collectConfigMapArtifact`.

internal/cmds/nodeservices/prepare.go
joaosa: Addressed in fb8fbde8: `serverOutputs` and `removeServerOutputs` around a `serverConfig` struct, with one shared `serverOutputNames` list so the set a server writes and the set an agent clears cannot drift.

node-guest-image/c8s/mkosi.extra/etc/systemd/system/attest-proxy.service
joaosa: The socket. `CAP_CHOWN` exists only so attest-proxy can chgrp the socket it creates at `--socket` to `--socket-gid` (65532), letting the non-root workloads in that group connect. Nothing under `/dev` is touched; the attester is reached over `--upstream` on loopback. Comment added in fb8fbde8.

cmd/c8s/install.go
joaosa: Addressed in 0598e3ac: now reads "regardless of the CPU TEE type".

cmd/c8s/node_services_linux.go
joaosa: Addressed in 0598e3ac

internal/helmchart/c8s/templates/cds.yaml
joaosa: No. RKE2 sets node-role.kubernetes.io/control-plane on servers only, so agents never match and this pins CDS to the server. It does assume the single server a baked image boots; fb8fbde8 says so in the template. Making the pin follow the signed role instead of the RKE2 role is now an issue

internal/helmchart/c8s/templates/router-deployment.yaml
joaosa: Right, and agreed it is fragile under a second server. Kept as a documented single-server assumption for this round. opened an issue

internal/cmds/getcert/run.go
joaosa: Agreed. Kept out of this PR; opened an issue for the sweep.

cmd/c8s/node_image.go
```diff
@@ -0,0 +1,384 @@
...
+		}
+	}
+	if cfg.outputDir == "" {
+		return fmt.Errorf("--output-dir is required")
+	}
+	return nil
+}
+
+// images maps build flags to the chart's authoritative image values.
+func (cfg nodeImageRenderConfig) images() []struct {
```
CluEleSsUK: using a named struct might save some space and mental energy reading this func

internal/cmds/nodeservices/prepare.go
```diff
@@ -0,0 +1,251 @@
...
+	pins, err := refvalues.Parse(data)
+	if err != nil {
+		return nil, fmt.Errorf("staged identity policy %s: %w", path, err)
+	}
+	family, err := teetypes.ParseFamily(platform)
+	if err != nil || pins.Family != family || len(pins.Images) == 0 || (serverOnly && len(pins.Images) != 1) {
+		return nil, fmt.Errorf("staged identity policy %s has invalid family or image count", path)
+	}
+	for _, pin := range pins.Images {
+		if len(pin.Anchor) == 0 || (family == teetypes.FamilyTDX && (len(pin.RTMRs[1]) == 0 || len(pin.RTMRs[2]) == 0)) {
```
CluEleSsUK: could extract to its own func for readability

internal/cmds/nodeservices/prepare.go
```diff
@@ -0,0 +1,251 @@
...
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	pins, err := refvalues.Parse(data)
+	if err != nil {
+		return nil, fmt.Errorf("staged identity policy %s: %w", path, err)
+	}
+	family, err := teetypes.ParseFamily(platform)
+	if err != nil || pins.Family != family || len(pins.Images) == 0 || (serverOnly && len(pins.Images) != 1) {
```
CluEleSsUK: could extract to its own func for readability

internal/cmds/nodeservices/prepare.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	return seed.Canonical()
+}
+
+// allowIdentical is used only for the two measured bootstrap sources. Signed
+// tenant workloads may never replace a component, even with identical content.
+func mergeWorkloads(base, extra *allowlist.Allowlist, allowIdentical bool) error {
+	for name, workload := range extra.Workloads {
+		if existing, exists := base.Workloads[name]; exists {
+			if allowIdentical {
```
CluEleSsUK: would stick error case first with `if !allowIdentical`

CluEleSsUK (approved):
a few nits but ship it!

cmd/c8s/node_image.go
joaosa: Addressed in ce6106a3: `nodeImageInput` with named fields.

internal/cmds/nodeservices/prepare.go
joaosa: Addressed in ce6106a3: `policyPinsComplete`.

internal/cmds/nodeservices/prepare.go
joaosa: Addressed in ce6106a3: `policyMatchesPlatform`.

internal/cmds/nodeservices/prepare.go
joaosa: Addressed in ce6106a3: the missing-workload and `!allowIdentical` cases now return first, so the identical-content comparison is the unnested tail.

PR #666, opened 2026-09-18
fix(nri-image-policy): classify state overlays from the boot scratch mapping

CluEleSsUK:
## Why

A disk-backed `emptyDir` was admitted on a synthetic node and denied on the real
node image, for the same pod and the same `exact` mount rule. The image runs a
read-only dm-verity root; the measured initrd mounts the encrypted scratch at
`/state` and gives each directory declared in `/usr/lib/confai/state.d` an
overlay with `upperdir=/state/N/upper`. `switch_root` then discards that mount
namespace, so the upperdir the kernel keeps printing in `/proc/self/mountinfo`
names a path that no longer resolves. `inspect()` recursed onto it, `Statfs`
failed, and every disk-backed `emptyDir` on `/var` came back `unknown` — denied.
`hasScratchSlave`, the line #652 fixed, sits downstream of that call and never
ran.

After this change a state overlay is classified from the scratch mapping
`scratch-enforce.service` recorded for the current boot, so the same pod is
admitted with storage `encrypted`.

## The change

An overlay is treated as the initrd's state overlay when the most specific mount
containing the source is a writable overlay **and** its mountpoint is exactly one
of the directories the measured image declares in `/usr/lib/confai/state.d`. For
those, storage comes from `/run/c8s/scratch-provenance.json` — re-verified live
through sysfs: the recorded device is a dm mapping named `scratch`, with a
`CRYPT-` UUID, whose device number and boot ID match, backed by a block device
whose serial is `confai-scratch`. Every other overlay keeps the existing
upperdir recursion, so nodes without `state.d` are unaffected.

- The provenance `device` field now builds a path, so it is rejected unless it is
  `major:minor`; it cannot name anything outside `/sys/dev/block`.
- `inspect()` resolves the source before the mountinfo lookup, because `Statfs`
  follows symlinks and the lexical prefix match did not.
- The scratch arm of `trustedEncryptedDevice` is factored out and shared, so both
  callers apply the identical three-fact contract.
- `.github/scripts/node-guest-image-invariants.sh` pins the confos contract this
  now depends on (`mount_state_backing /state`, `upperdir=/state/$2/upper`): if a
  future initrd backs some overlay differently, CI fails instead of the
  classifier silently over-admitting.

## Alternatives rejected

- **Rewriting `/state/...` or `/sysroot/...` to a live-root path.** There is no
  mapping to rewrite to; `switch_root` detaches the subtree, so nothing under
  `/state` is reachable by any path. A guessed prefix would turn a host-shaped
  mountinfo string into a live-root lookup.
- **Classifying from `st.Dev` of the overlay's upper layer.** The inspected
  source is a directory, and overlayfs reports its own pseudo device for
  directories unconditionally (with `xino`, for every object). That device has no
  `/sys/dev/block` entry, so the slave walk has nothing to start from.
- **Trusting any overlay whose upperdir does not resolve.** That is the
  fail-open shape: a mount with a deliberately absent upperdir would classify
  itself as encrypted.
- **Matching the overlay's `lowerdir` against `/sysroot` + its mountpoint.**
  Stronger anti-forgery, but it depends on how the kernel formats mount options
  (6.7 prints `lowerdir+=` for fsconfig-mounted overlays), and a format change
  would silently reproduce exactly this outage. The mountpoint is structural
  mountinfo data instead.
- **Recording the state overlays' mount IDs in the provenance file.**
  `scratch-enforce.service` runs after `switch_root` and cannot observe the
  linkage either, so it would record the same inference — at the cost of a
  schema change to an interface the node image and the plugin ship separately.

## Fail-closed

The decision still bottoms out in `trustedScratchProvenance` + `hasScratchSlave`
against sysfs; a forged provenance file alone proves nothing. The narrowing that
changed is which mountinfo field is load-bearing: previously the `upperdir`
option, now the mountpoint. Both are equally kernel-generated, and both require
an adversary who can already mount in the node's mount namespace — for the
mountpoint they would have to remount `/var` itself on a live node.

What is genuinely weaker: the upper layer's device number is no longer confirmed
per mount, because the kernel exposes no way to do that once the layer is
unreachable. It is replaced by the node-wide fact that the measured initrd backs
every state overlay from the one scratch mapping, which is what the new CI pin
holds in place.

## Tests

`TestOverlayStorageClassifiesInitrdStateOverlay` is the regression test. The gap
the existing tests cannot close by construction is that they record an upperdir
they just created, so it always exists; this one records `/state/1/upper`,
asserts it does not resolve (`inspect` on it returns `unknown` — the old
behaviour), and then asserts the overlay classifies as `encrypted` anyway.
`TestOverlayStorageFailsClosed` covers twelve ways it must not: undeclared or
commented-out mountpoint, a declaration that is only an ancestor, no `state.d`
at all, a read-only overlay, missing provenance, a previous boot's ID, a device
number that tries to escape sysfs, a device number disagreeing with sysfs, a
non-crypt target, a mapping merely named `scratch` elsewhere, and a wrong
backing serial.

## Review focus

- `overlayStorage` in `internal/cmds/nri-image-policy/mount_storage_linux.go`:
  is "mountpoint is a declared state dir + boot scratch mapping" a proof you
  accept, given the per-mount device check cannot be reconstructed?
- The `state.d` parser is now the third implementation of that format (confos
  initrd, the invariant lint, this).
- `inspect()` resolving symlinks changes every storage classification, not just
  the overlay path.

## Unverified

There is no cluster or node image available here, so nothing below Go tests has
run on hardware. Confirm on a SEV-SNP node-CVM that a disk-backed `emptyDir`
under an `exact` rule is admitted, and that `/usr/lib/confai/state.d/*.conf` on
the booted image lists `var` (the mountinfo evidence in the report implies it —
`/state/1` is the first declared directory — but it was not read directly).


internal/cmds/nri-image-policy/mount_storage_linux.go
```diff
@@ -19,23 +19,29 @@ import (
...
 	provenanceFile string
 	bootIDFile     string
 }
 
 func newMountStorageInspector() mountStorageInspector {
 	return linuxStorageInspector{
 		sysDevBlock: "/sys/dev/block", sysClassBlock: "/sys/class/block",
-		mountInfo: "/proc/self/mountinfo", provenanceFile: "/run/c8s/scratch-provenance.json",
-		bootIDFile: "/proc/sys/kernel/random/boot_id",
+		mountInfo: "/proc/self/mountinfo", stateDir: "/usr/lib/confai/state.d",
```
CluEleSsUK: format these to one-per-line

internal/cmds/nri-image-policy/mount_storage_linux.go
```diff
@@ -106,6 +105,87 @@ func (i linuxStorageInspector) trustedEncryptedDevice(device string, seen map[st
 	})
 }
 
+// trustedScratchMapping accepts the measured initrd's scratch contract, which
+// needs all three facts: exact mapper name, a crypt target UUID, and a backing
+// virtio device whose exact serial is the launch contract.
+func (i linuxStorageInspector) trustedScratchMapping(device string) bool {
```
CluEleSsUK: isTrustedScratchMapping

internal/cmds/nri-image-policy/mount_storage_linux.go
```diff
@@ -106,6 +105,87 @@ func (i linuxStorageInspector) trustedEncryptedDevice(device string, seen map[st
...
+		return allowlist.MountEncrypted
+	}
+	return i.inspect(upper, seen)
+}
+
+// declaredStateDir reports whether mountpoint is one of the directories the
+// measured image declares for a writable state overlay, parsed the way the
+// initrd parses them: one path per line, blank lines and # comments skipped,
+// a leading slash tolerated.
+func (i linuxStorageInspector) declaredStateDir(mountpoint string) bool {
```
CluEleSsUK: isDeclaredStateDir

Issue #669, opened 2026-09-18
GPU attestation must run on every GPU appearance, not only at boot; refuse non-CC GPUs (hotplug/reattach gap)


aamirrasheed:
## Summary

The GPU attestation gate runs one time, at boot. It is a `Type=oneshot` unit with `RemainAfterExit=yes`. The requirement is stronger: a GPU must attach to a c8s node in CC mode only, on every appearance, no matter when or how many times that appearance happens.

Two gaps follow. First, a GPU that appears after boot gets no check. Second, the gate passes when it sees no GPU at all, so a hidden GPU also gets no check.

## What we observed

We run `gpu-cc-enforce.sh` and its systemd unit on our staging cluster (CPU-only TDX nodes), at tag `v0.26.1`. On a node with no GPU, the script exits 0 before any HTTP call runs. This is the designed no-op path. The same code path is the gap on a GPU node: a hidden GPU produces the same "nothing to enforce" exit.

## Evidence (v0.26.1)

`node-guest-image/c8s/mkosi.extra/usr/local/bin/gpu-cc-enforce.sh`:
- Lines 33-40: the script scans PCI devices for vendor `0x10de` (NVIDIA). If it finds none, line 41-44 prints "no NVIDIA GPU present; nothing to enforce" and exits 0.
- Lines 73-76: the script reads `gpu_count` from `nvidia-smi --query-gpu=uuid`. This value becomes the script's own expected count.
- Lines 115-126: the script compares the attested GPU count against `$want`, which is `gpu_count` from the same guest. The check never compares against a count from outside the guest.

`node-guest-image/c8s/mkosi.extra/etc/systemd/system/gpu-cc-enforce.service`:
- Line 12: `Before=nvidia-cc-ready.service rke2-server.service rke2-agent.service`.
- Line 15-16: `Type=oneshot` and `RemainAfterExit=yes`. The unit runs one time per boot and then stays active.
- Line 23: `FailureAction=poweroff-force`.
- Line 33: `RequiredBy=nvidia-cc-ready.service rke2-server.service rke2-agent.service`.

This unit was added by PR #648 (commit `29dbf4d1`).

No GPU attestation result reaches `cds-attest`, `cds`, node discovery, or receipts. A search for "gpu" in `internal/cmds/cds` and `internal/cmds/cdsattest` returns no matches. The script and its systemd unit are in the verity root and are covered by MRTD/RTMR1/RTMR2, per the node-guest-image README.

[...]

PR #670, opened 2026-09-18
fix(node-image): check GPU CC mode per device in gpu-cc-enforce

aamirrasheed:
## Summary
gpu-cc-enforce.sh reads CC mode from one aggregate nvidia-smi line. It passes when that one line says on. This change queries every GPU by index. It requires each GPU's own line to say on. The logic becomes auditable from the script alone.

## Why
This is not a proven vulnerability. The single line comes from NVML's system-level CC state. The closed driver computes this value internally. No NVIDIA document states how the value is derived when GPUs have independent CC modes set on the host. Independent modes can mean an AND rule, a first-GPU-only rule, or a driver refusal of a mixed set. The script's own comment says the driver prints "one line per GPU with -i, one aggregate line without" (`node-guest-image/c8s/mkosi.extra/usr/local/bin/gpu-cc-enforce.sh:60-61`). A reviewer of the open image cannot resolve this question from source code alone. The per-device query removes the dependency on undocumented driver behavior. The check then says exactly "every enumerated GPU reports CC on". Anyone can verify this claim by reading the script. (Refs #669.)

## The check today, step by step
1. Run `nvidia-smi conf-compute -f`. Retry until the driver answers, or until `DRIVER_TIMEOUT` expires. A timeout is a failure (`gpu-cc-enforce.sh:54-58`).
2. Keep only lines that match `cc (feature|status)\s*:`, case-insensitive. Driver version 595 prints "CC status". Older drivers print "CC feature" (`gpu-cc-enforce.sh:60`, prior behavior).
3. Require at least one kept line. If there are none, fail with "cannot parse" (prior behavior).
4. Fail if any kept line does not end in "on" or "enabled" (prior behavior).
5. These steps stay the same after this change: query `--query-gpu=uuid`, require at least one UUID, send a POST request to the attestation API with `nvidia_gpu_required`, require the attested GPU count to equal the enumerated GPU count, and require signature_valid, nonce_binding_ok, report_data_match, and overall_ok (`gpu-cc-enforce.sh:82-126`).

Without `-i`, step 1 returns one line. Step 4 then tests one value for the whole node.

## The check after this change, step by step
1. Enumerate GPUs as pairs of index and UUID (`gpu-cc-enforce.sh:64-66`).
2. For each GPU, run step 1 above with `-i <index>` (`gpu-cc-enforce.sh:70-71`).
3. Apply steps 2 through 4 above to that GPU's own output (`gpu-cc-enforce.sh:72-76`).
4. On failure, name the GPU's index and its UUID (`gpu-cc-enforce.sh:73,75`).
5. Step 5 above stays the same.

The new per-device loop (`gpu-cc-enforce.sh:64-79`):
```sh
cc_gpu_list=$(nvidia-smi --query-gpu=index,uuid --format=csv,noheader 2>&1) \
    || fail "nvidia-smi --query-gpu failed: $cc_gpu_list"
[ -n "$cc_gpu_list" ] || fail "nvidia-smi enumerates no GPU despite CC mode on"
while read -r cc_gpu_line; do
    cc_gpu_index=${cc_gpu_line%%,*}
    cc_gpu_uuid=${cc_gpu_line#*, }
    cc_status=$(nvidia-smi conf-compute -f -i "$cc_gpu_index" 2>&1) \
        || fail "nvidia-smi conf-compute -f -i $cc_gpu_index failed: $cc_status"
    cc_line=$(printf '%s\n' "$cc_status" | grep -iE 'cc (feature|status)[[:space:]]*:' || true)
    [ -n "$cc_line" ] || fail "cannot parse CC status for GPU $cc_gpu_index ($cc_gpu_uuid): $cc_status"
    if printf '%s\n' "$cc_line" | grep -qviE ':[[:space:]]*(on|enabled)[[:space:]]*$'; then
        fail "GPU $cc_gpu_index ($cc_gpu_uuid) not in CC mode ($cc_line). Enable GPU CC mode on the host (nvidia_gpu_tools.py --set-cc-mode=on)"
    fi
done <<EOF
$cc_gpu_list
EOF
```

## Test
`node-guest-image/tests/gpu-cc-enforce-test.sh` stubs `nvidia-smi` to answer per GPU index. The mixed case sets GPU 0 to "CC status: ON" and GPU 1 to "CC status: OFF" (`gpu-cc-enforce-test.sh:137-140`). This case must fail and must name the failing GPU, checked as `stderr_has "GPU 1 (GPU-1)"`. The all-on case (`gpu-cc-enforce-test.sh:128-130`) must pass. The full suite reports PASS: 32, FAIL: 0. `shellcheck -x -P SCRIPTDIR` on both files reports only pre-existing SC2016 info notices (single-quoted heredoc-generator lines), unchanged from before this change.

## Out of scope
Expected GPU count (issue #669) is reporting only. It is not part of this change. Hotplug is already off in the guest kernel. `CONFIG_HOTPLUG_PCI` is not set. Every GPU is present at boot.

## Open question for reviewers
How does the driver compute the aggregate line for a mixed set of CC modes? It may use an AND rule, or the driver may refuse a mixed set outright. If either is true, this change is for auditability only. If neither is true, this change also closes a real gap. We will capture the aggregate output on a real multi-GPU node with one GPU set to CC-off, and add that evidence here.


aamirrasheed:
withdrawn pending evidence

AmeanAsad (approved):

aamirrasheed:

aamirrasheed:
recheck

CluEleSsUK (approved):

